Issue #13, opened 2017-10-23
Macros or attributes to prevent repetition of impl ResponseType


superlou:
I've noticed that I have a lot of repetition of `impl ResponseType<...` since a new one is required for each signal.  To DRY my code, I've defined the following macro:

```
macro_rules! response_type {
    ($Actor:ty, $Msg:ty, $item:ty, $err:ty) => (
        impl ResponseType<$Msg> for $Actor {
            type Item = $item;
            type Error = $err;
        }
    );

    ($Actor:ty, $Msg:ty, $item:ty) => (
        impl ResponseType<$Msg> for $Actor {
            type Item = $item;
            type Error = ();
        }
    );

    ($Actor:ty, $Msg:ty) => (
        impl ResponseType<$Msg> for $Actor {
            type Item = ();
            type Error = ();
        }
    );
}
```

This seems to work, though there may be a more elegant approach.  In a perfect world, it would be great to use an attribute as in the following example:

```
#[response_type(MyStatusEnum)]
impl Handler<PrivMsg> for Engine {
    fn handle(&mut self, msg: PrivMsg, ctx: &mut Context<Self>) -> Response<Self, PrivMsg> {
        println!("Source: {}, Target: {}, Text: {}", &msg.source, &msg.target, &msg.text);
        MyStatusEnum::Ok
    }
}
```

Is there a technique I missed that would make all of the above superfluous? Anyone else running into this repetition?

fafhrd91:
1. `#[response_type(MyStatusEnum)]` is possible with proc_macro. but this works only on nightly.
also it shoulde be possible to destructure message and use message attributes as 
`handle` method parameters combined with impl Trait, it could be very nice

something like:
```rust
impl Engine {
    #[handler(PrivMsg)]
    fn do_thomething(&mut self, source: Source, target: Target, text: Text) -> Result<PrivMsgResponse> {
        println!("Source: {}, Target: {}, Text: {}", &msg.source, &msg.target, &msg.text);
        MyStatusEnum::Ok
    }
}
```

2. I've been thinking about `ResponseType`, I think it was wrong decision to use impl ResponseType on Actor, it is property of message itself. I changed master, message has to impl ResponseType.

superlou:
On point 1, that seems to be a more typical pattern on frameworks I've seen in Rust, notably Rocket.  Callback-ish code is just a function, and the framework gets the information it needs from an attribute.

[...]

PR #18, opened 2017-11-16
Add Proc Macros

callym:
Adds support for the following custom derives, as discussed in #13 
- `[derive(Message)]`
    - `#[MessageResult(type)]`
    - `#[MessageError(type)]`
    - `#[Message(result_type, error_type)]` (where `error_type` is optional)
```rust
#[derive(Message)]
#[MessageResult(usize)]
struct Sum(usize, usize);

#[derive(Actor)]
struct SumActor;
```

Which expands to:
```rust
impl ResponseType for Sum {
    type Item = usize;
    type Error = ();
}

impl Actor for SumActor {
    type Context = Context<Self>;
}
```

Adds support for a `#[handler]` macro when used with the `handler` feature. (Requires the `proc_macro` feature and therefore nightly)
```rust
#![feature(proc_macro)]

#[handler]
impl SumActor {
    #[handle(Sum)]
    fn sum(&mut self, message: Sum, _: &mut Context<Self>) -> Response<Self, Sum> {
        println!("{}", message.0 + message.1);
        Self::reply(message.0 + message.1)
    }
}
```

Which expands to:
```rust
#![feature(proc_macro)]

impl SumActor {
    fn sum(&mut self, message: Sum, _: &mut Context<Self>) -> Response<Self, Sum> {
        println!("{}", message.0 + message.1);
        Self::reply(message.0 + message.1)
    }
}

impl Handler<Sum> for SumActor {
    fn handle(&mut self, message: Sum, context: &mut Context<Self>) -> Response<Self, Sum> {
        self.sum(message, context)
    }
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,9 +46,6 @@ crossbeam = "0.3"
 libc = { version = "0.2", optional = true }
 uuid = { version = "0.5", features = ["v4"] }
 
-# derive
-actix_derive = { path = "actix_derive", version = "0.1.0" }
-
 [dependencies.tokio-signal]
 version = "0.1"
 optional = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ signal = ["tokio-signal", "libc"]
 dns = ["libc"]
 
 [workspace]
-members = ["examples/chat"]
+members = ["examples/chat", "actix_derive"]
 
 [dependencies]
 # tokio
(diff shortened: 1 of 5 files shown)


callym:
I've pushed the `#[Message(response, error)]` macro to https://github.com/actix/actix-derive, and published it on crates.io

I also opened issues to discuss the `actor` custom derive, and `handler` proc macro.

PR #22, opened 2017-11-21
Fix bug in SystemRegistry

### LINKED ISSUE #21: SystemRegistry bug
Hi,

I think I found a bug in the SystemRegistry, it does not actually write to the hashmap when it creates a service, thus every time a service is looked up, a new one is created.

I have attached a patched version of this function below:
```rust
    pub fn get<A: SystemService + Actor<Context=Context<A>>>(&self) -> SyncAddress<A> {
        if let Ok(hm) = self.registry.lock() {
            if let Some(addr) = hm.borrow().get(&TypeId::of::<A>()) {
                match addr.downcast_ref::<SyncAddress<A>>() {
                    Some(addr) =>{
                        return addr.clone()},
                    None =>
                        error!("Got unknown value: {:?}", addr),
                }
            }
            let addr = Supervisor::start_in(Arbiter::system_arbiter(), false, |ctx| {
                let mut act = A::default();
                act.service_started(ctx);
                act
            });
            hm.borrow_mut().insert(TypeId::of::<A>(), Box::new(addr.clone().unwrap()));
            return addr.expect("System is dead")
        }
        panic!("System registry lock is poisoned");
    }
```

kingoflolz:
fixes #21

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -146,19 +146,19 @@ impl SystemRegistry {
         if let Ok(hm) = self.registry.lock() {
             if let Some(addr) = hm.borrow().get(&TypeId::of::<A>()) {
                 match addr.downcast_ref::<SyncAddress<A>>() {
-                    Some(addr) =>
-                        return addr.clone(),
+                    Some(addr) =>{
+                        return addr.clone()},
                     None =>
                         error!("Got unknown value: {:?}", addr),
                 }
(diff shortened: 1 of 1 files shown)


fafhrd91:
0.3.3 version is released

kingoflolz:
thanks for the quick release!

PR #25, opened 2017-11-24
Allow returning Box<ActorFuture> in ActorFutures

kingoflolz:
This allows for non tail recursive futures, such as [here](https://github.com/kingoflolz/routing-actor/blob/master/src/dht/service.rs#L100)

### DIFF
--- a/changes.md
+++ b/changes.md
@@ -1,5 +1,9 @@
 # CHANGES
 
+## Master
+
+* Allow returning Box<ActorFuture> in ActorFutures for non tail recursive futures
+
 ## 0.3.2 (2017-11-06)
 
 * Disable `signal` feature by default
--- a/src/fut/mod.rs
+++ b/src/fut/mod.rs
@@ -189,6 +189,20 @@ impl<F: ActorFuture> IntoActorFuture for F {
     }
 }
 
+impl<F: ActorFuture + ?Sized> ActorFuture for Box<F> {
+    type Item = F::Item;
+    type Error = F::Error;
+    type Actor = F::Actor;
+
+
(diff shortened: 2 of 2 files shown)


fafhrd91:
Do you expect more changes to actix? Or I can make new release

fafhrd91:
btw you use actix for some time now, would you provide feedback? any annoyances, ergonomic problems?

kingoflolz:
Not sure why I had to do 2 merge commits, but updated changes.md

I might do some more changes, so you don't need to release straight away.

re Feedback: Actix is really nice and ergonomic, really the only gripe I have with it is about performance, that I don't think that futures/tokio with per actor queues and polling will achieve very good cache util and performance. Something like Citybound's Kay with per "chunk" (2mbish of actors) queues which can be polled at once can achieve higher performance. Although I guess my use case (thousands-millions of actors) isn't the typical use case for this library. (also another idea is to have chunked per thread outboxes which fill up, and then send their outboxes to the inbox, to reduce the amount of sync overhead, see: crossbeam::sync::SegQueue)

An easy way to improve this is just to switch to crossbeam-channel instead of std channels, which is approx 5x faster in MPSC cases.

fafhrd91:
that is interesting, but I am not sure any actor system can handle thousands millions of actors, you probably need to build some custom scheduling system, and use actix as an interface. 

btw there actix has https://github.com/actix/actix/blob/master/src/sync.rs actors, sync actors more like cpu bound workers, I use `crossbeam::sync::MsQueue` and queue. you can build something similar.

fafhrd91:
I am planing to make another release. @kingoflolz are you planing to do anything else with this PR?

fafhrd91:
closing. this impl does not work on stable at the moment.

PR #27, opened 2017-12-10
Add convenience impl for boxed Subscriber objects.

proman21:
Hi, loving using this library at the moment. I just thought I'd add some convenience to using boxed `Subscriber`s so that they can be cloned using `.clone()`.

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -50,13 +50,18 @@ impl<A> ActorAddress<A, ()> for A where A: Actor {
 }
 
 pub trait Subscriber<M: 'static> {
-
     /// Send buffered message
     fn send(&self, msg: M) -> Result<(), M>;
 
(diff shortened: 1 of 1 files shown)


src/address.rs
```diff
@@ -50,13 +50,18 @@ impl<A> ActorAddress<A, ()> for A where A: Actor {
...
 
 pub trait Subscriber<M: 'static> {
-
     /// Send buffered message
     fn send(&self, msg: M) -> Result<(), M>;
 
-    /// Clone subscriber
-    fn subscriber(&self) -> Box<Subscriber<M>>;
+    /// Clone the current subscriber
+    fn boxed_clone(&self) -> Box<Subscriber<M>>;
```
fafhrd91: Let's name it just `boxed`

PR #36, opened 2018-01-07
fix pop wrong item when cacel in handle

woodgear:
when you use cancel_future in handle it will sub the items len so the items.pop() will pop prevent item.
you could use this demo to see the different
or maybe i should not use self notify like this
PS: I do not fully understand actix's logic. so i just add patch fix that
~~~
// [dependencies]
// time = "0.1.39"
// actix = "0.4.0"
extern crate actix;
extern crate time;
use actix::prelude::*;
use actix::prelude::actix::SpawnHandle;
use std::time::Duration;
use std::thread::sleep;
use std::thread;

#[derive(Debug)]
enum Event {
    Update(String),
    Notify,
}
impl ResponseType for Event
{
    type Item = ();
    type Error = ();
}

struct ActorDemo {
    data:Option<String>,
    notify_handle: Option<SpawnHandle>
}

impl Actor for ActorDemo {
    type Context = Context<Self>;
    fn started(&mut self, _: &mut Self::Context) {
    }
}

impl Handler<Event> for ActorDemo {
    type Result = Result<(),()>;
    fn handle(&mut self, event: Event, ctx: &mut Context<Self>) -> Self::Result {
        println!("event {:?}",event);
        match event {
            Event::Update(data) => {
                println!("update ===================================> {}",data);
                self.data=Some(data);
            },
            Event::Notify => {
                if let Some(ref data) = self.data {
                    println!("notify ===========================>{}",data);
                }
            },
        }

        if let Some(notify_handle) = self.notify_handle {
            println!("cacel handle {:?}",notify_handle);
            ctx.cancel_future(notify_handle);
        }

        let notify_handle = ctx.notify_later(Event::Notify, Duration::from_secs(3));
        println!("set notify_handle  {:?}",notify_handle);
        self.notify_handle = Some(notify_handle);
        Ok(())
    }
}

fn main() {
    let system = System::new("test");
    let actor = ActorDemo {
        notify_handle: None,
        data:None,
    };
    let address:SyncAddress<_> = actor.start();

    thread::spawn(move || {
        loop {
            address.send(Event::Update("test".to_owned()));
            sleep(Duration::from_secs(10000));
        }
    });

    
    system.run();

}

~~~

### DIFF
--- a/src/contextcells.rs
+++ b/src/contextcells.rs
@@ -193,6 +193,10 @@ impl<A> ActorItemsCell<A> where A: Actor, A::Context: AsyncContext<A>
         }
         false
     }
+    
+    fn find_index_by_handle(&self, handle:SpawnHandle)-> Option<usize> {
+        self.items.iter().position(|ref x| x.0 == handle)
+    }
 
     pub fn poll(&mut self, act: &mut A, ctx: &mut A::Context, stop: bool) -> ContextCellResult {
         loop {
@@ -201,6 +205,7 @@ impl<A> ActorItemsCell<A> where A: Actor, A::Context: AsyncContext<A>
             let mut not_ready = true;
 
             while idx < len {
+                let handle = self.items[idx].0;
                 let drop = match self.items[idx].1.poll(act, ctx) {
                     Ok(val) => match val {
                         Async::Ready(_) => {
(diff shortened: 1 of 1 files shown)


fafhrd91:
thanks for PR. 

I think better would be remove items from vec only inside poll.

woodgear:
What you mean about that? refactor the implement of cacel_future?

fafhrd91:
I pushed fix in b651491

woodgear:
that cool

woodgear:
when will you make a release? i need this patch.
or i could use git repo

fafhrd91:
0.4.2 is published

PR #46, opened 2018-02-13
spelling check

andreevlex:


src/actor.rs
```diff
@@ -47,10 +47,10 @@ use utils::TimerFunc;
 ///
 /// * `Context::stop` get called by actor itself
 /// * all addresses to the actor get dropped
-/// * no evented objects are registered in context.
+/// * no events objects are registered in context.
```
fafhrd91: this is not right.

src/actor.rs
```diff
@@ -47,10 +47,10 @@ use utils::TimerFunc;
...
 ///
 /// Actor could restore from `stopping` state to `running` state by creating new
-/// address or adding evented object, like future or stream, in `Actor::stopping` method.
+/// address or adding event object, like future or stream, in `Actor::stopping` method.
```
fafhrd91: same

fafhrd91:
thanks!

fafhrd91:
we probably should create file with words to exclude

PR #55, opened 2018-03-05
Message link was pointing to Actor trait

sbditto85:
The link to the Message trait was pointed towards the Actor trait, changed to point towards the Message trait.


fafhrd91:
thanks!

PR #56, opened 2018-03-06
Created Constructor to allow TcpConnect timeout to be set

glademiller:


src/actors/resolver.rs
```diff
@@ -74,14 +79,43 @@ impl Message for Resolve {
 pub struct Connect {
     name: String,
     port: Option<u16>,
+    timeout: Option<Duration>,
```
fafhrd91: let's just  use `Duration`

src/actors/resolver.rs
```diff
@@ -74,14 +79,43 @@ impl Message for Resolve {
...
     pub fn host_and_port<T: AsRef<str>>(host: T, port: u16) -> Connect {
-        Connect{name: host.as_ref().to_owned(), port: Some(port)}
+        Connect {
+            name: host.as_ref().to_owned(),
+            port: Some(port),
+            timeout: None,
+        }
+    }
+
+    pub fn host_and_timeout<T: AsRef<str>>(host: T, timeout: Duration) -> Connect {
```
fafhrd91: let's use builder pattern here. 

```rust
pub fn timeout(mut self, timeout: Duration) -> Connect {
    self.timeout = timeout;
    self
}
```

src/actors/resolver.rs
```diff
@@ -74,14 +79,43 @@ impl Message for Resolve {
...
+        Connect {
+            name: host.as_ref().to_owned(),
+            port: None,
+            timeout: Some(timeout),
+        }
+    }
+
+    pub fn host_and_port_and_timeout<T: AsRef<str>>(
```
fafhrd91: we dont need this method with builder

src/actors/resolver.rs
```diff
@@ -244,9 +287,11 @@ impl ActorFuture for Resolver {
     type Error = ConnectorError;
     type Actor = Connector;
 
-    fn poll(&mut self, _: &mut Connector, _: &mut Context<Connector>)
-            -> Poll<Self::Item, Self::Error>
-    {
+    fn poll(
```
fafhrd91: i dont like how this formatting looks

glademiller: Oops yep I'll turn off autoformatting

fafhrd91:
Could you also update Changes.md

PR #62, opened 2018-03-30
Guide: improve wording & style

bwasty:
Same as for `actix-web` - I tried to improve the language a bit while reading through.


fafhrd91:
great! thanks!

PR #71, opened 2018-04-28
add set function to registry and system registry

kingoflolz:


fafhrd91:
Do we need .set method for normal registry?

PR #78, opened 2018-05-15
Ignore rls dir at project root dir. This dir created by latest version of rls tool.

zengsai:


zengsai:
yes, It has been fixed.

PR #90, opened 2018-06-07
Update README.md: fix doc links for Unsync + Syn

ms-ati:
I was clicking through, and observed that the existing links were broken.


fafhrd91:
thanks!

PR #91, opened 2018-06-07
Remove stutter by aliasing `registry::Registry`

notriddle:


fafhrd91:
thanks!

this is fixed

PR #93, opened 2018-06-14
Add run_until_complete to run single future

DoumanAsh:
I'd like to return option to run single future without a need for manual stopping

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -148,11 +148,24 @@ impl System {
         }
     }
 
-    fn create_runtime<F>(name: &str, pool_size: usize, f: F) -> (Runtime, Receiver<i32>)
+    ///Runs future with minimal runtime.
+    ///
+    ///Once future is complete runtime shall be shutdown.
+    ///
+    ///NOTE: Actual System is not created so it cannot be stopped by means of [stop](struct.System.html#method.stop)
+    ///Instead it is stopped as soon as future is resolved.
+    ///Due to that [current](struct.System.html#method.current) shall panic on attempt to access it.
+    pub fn run_until_complete<I, E, F>(f: F) -> Result<I, E>
     where
-        F: FnOnce() + Send + 'static,
+        I: Send + 'static,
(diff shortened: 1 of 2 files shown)


fafhrd91:
tokio runtime provides [block_on](https://docs.rs/tokio/0.1.7/tokio/runtime/struct.Runtime.html#method.block_on)

fafhrd91:
I mean, let’s use block_on instead of combinator

fafhrd91:
@DoumanAsh but what is the point then for `run_until_completed`? you can use just block_on directly.

btw should we use `block_on` name for consistency?

DoumanAsh:
@fafhrd91 Since Actix wraps tokio user should not have need to be exposed to settings of tokio-runtime as actix should choose optimal configuration of runtime for him.

While I know that tokio will be started automatically too, but its defaults are different of these Actix uses.
So I think we should continue to take care of it for user.

If you agree I can rename method to `block_on`

fafhrd91:
I think if we implement this function, runtime should have system. Otherwise it is just too similar to tokio block_on

DoumanAsh:
If we going this way, then I suspect we may not be able to return value of future so signature will require `Future<Item=(), Error=()>` are you ok with that?

I'll look into it more detailed later on

fafhrd91:
Why? We should just use SystemRunner, then we can use same runtime, it contains system arbiter

DoumanAsh:
If we are going to start system, then user will have a way to stop it.
Which means we should decide whether we allow pre-mature end of future or not.
If we allow it then we cannot always return result.
If we do not allow, then I do not see much point in creating system at all.

I guess as you said we might suggest user to use `tokio::Runtimer.block_on` in this case

fafhrd91:
let's do the same as `run` method, just let wait future instead of stop

fafhrd91:
i restored old run_until_complete method

DoumanAsh:
Oh I see thank you

Issue #84, opened 2018-05-22
How to use an AsyncWriter with an actor?


alanhoff:
Hello folks, I have a `WriteHalf<TcpStream>` but I'm not sure how to attach it to an `Actor`. I wrote a [small prototype](https://gist.github.com/alanhoff/ed8f38afb3d80e0fa757eca339bdcd77) with an actor that reads from a `ReadHalf<TcpStream>` and writes to a `WriteHalf<TcpStream>`, basically an echo server. The problem is that it's not writing/flushing immediately:

<img width="198" alt="screen shot 2018-05-22 at 15 11 41" src="https://user-images.githubusercontent.com/1044223/40381666-b6eb1dbc-5dd2-11e8-8033-faca10bb09ac.png">

The desired output would be:

```
> hello\r\n
< hello\r\n
> world\r\n
< world\r\n
```

But the server is echoing the first line only when I send a second one.

Thanks!

fafhrd91:
you can set buffer size to 0, http://actix.rs/actix/actix/io/struct.FramedWrite.html#method.set_buffer_capacity

i am planing to revisit this api. write part is not in good shape.

alanhoff:
I'll reopen bc I think this might be a bug. Even using `set_buffer_capacity` didn't worked. Could it be related to the to the tokio version I'm using (0.1.6)? My example is here:

* https://github.com/alanhoff/rust-example-actix

fafhrd91:
version of tokio should not matter, I will check

alanhoff:
I made some kind of progress: it seems like using [`set_send_buffer_size(1)`](https://docs.rs/tokio/0.1.6/tokio/net/struct.TcpStream.html#method.set_send_buffer_size) kinda helps, when using socat the first command still buffers but the rest is sent immediately.

```
> hello\r\n
> world\r\n
< hello\r\n
< world\r\n
> hello\r\n
< hello\r\n
> world\r\n
< world\r\n
```

fafhrd91:
sorry, I am busy at the moment. I'll check this bug over weekend

alanhoff:
So it turns out it wasn't a bug, my `Stream` implementation was wrong. I pushed a working version to the same repository. Thx for the patience!

habnabit:
For what it's worth, #97 might be related.

PR #97, opened 2018-06-19
Notify the io::InnerWriter's task if necessary

habnabit:
>Unless the task is notified, the buffer isn't always flushed out after
every write. The actor's poll method would return NotReady when the
buffer empties without setting up a way to be notified when the buffer
was written to again.

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -4,7 +4,7 @@ use std::marker::PhantomData;
 use std::rc::Rc;
 
 use bytes::BytesMut;
-use futures::{Async, Poll};
+use futures::{Async, Poll, task};
 use tokio_io::codec::Encoder;
 use tokio_io::AsyncWrite;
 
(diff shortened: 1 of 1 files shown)


alanhoff:
This fixes the problem with #84, not sure why tests are failing on AppVeyor :/

DoumanAsh:
Well tests are passing on my machine.

@fafhrd91 should we merge it?

fafhrd91:
Ok

habnabit:
I'm not sure what, if anything, would be useful for me to do here. It seems like CI is broken in some way?

I suppose adding a functional test or two would be useful here. I'm not sure how to unit test task notification.

DoumanAsh:
@fafhrd91 Since nightly is currently broken, can you merge it without Travis check? It seems unnecessary to wait for it to get fixed

fafhrd91:
Done

fafhrd91:
@DoumanAsh you have permission to force push. Don’t overuse it :)

PR #99, opened 2018-06-20
AddressSender to use AtomicBool to allow Addr to be Sync/Send

DoumanAsh:
As subj states it makes `maybe_parked` wrapped into Atomic so that `AddressSender` and consequently `Addr` would be `Send` and `Sync` and it would allow us to store it in `actix-web` state


DoumanAsh:
P.s. I wonder if you meant here `Arc<Cell<bool>>`

DoumanAsh:
Oh Ok, i'll make PR

PR #100, opened 2018-06-20
libc should used on unix only

DoumanAsh:


DoumanAsh:
@fafhrd91 

Well currently `actix` itself doesn't use it for windows so I i think there is no need to link it on windows.
We can always return it to build on windows ones there is a need

fafhrd91:
Oh, thought this is actix-web

DoumanAsh:
np, I checked both build and tests on windows so there is no dependency on removed `libc`

PR #107, opened 2018-07-03
Implemented Eq/PartialEq for Addr.

cerceris:


src/address/mod.rs
```diff
@@ -132,6 +132,14 @@ impl<A: Actor> Clone for Addr<A> {
     }
 }
 
+impl<A: Actor> PartialEq for Addr<A> {
```
DoumanAsh: I think you could just use `#[derive(PartialEq, Eq)]` since the only field is `AddressSender`

fafhrd91:
could you also add entry to changes.md

src/address/mod.rs
cerceris: OK. I thought an explicit implementation would be more flexible. I'll change to `derive`.

DoumanAsh: That's understandable but one thing is that if, in future, we'll add something to `Addr` that is not itself `PartialEq` we would see it immediately(due to compile error) and consider whether we need to take it into account or not.

DoumanAsh: Hmmm that's strange actually, I guess I could check it myself later on, it is not an big issue

cerceris: Maybe some bounds are needed to be specified for `tx`, I don't know. I leave the explicit implementation for a while.

fafhrd91: I think derive takes into account `A`

cerceris:
Updated CHANGES.md.

src/address/mod.rs
cerceris: >> That's understandable but one thing is that if, in future, we'll add something to Addr that is not itself PartialEq we would see it immediately(due to compile error) and consider whether we need to take it into account or not.

It might be even worse if the newly-added field *does* implement PartialEq. This would break the address comparison semantics: two addresses are equal if and only if they refer to the same receiver.

DoumanAsh:
Thank you

PR #109, opened 2018-07-03
Implemented Hash for Addr.

cerceris:


fafhrd91:
We should not use unsafe.

cerceris:
It seems that Arc::into_raw/Arc::from_raw is an "officially approved" pattern.
https://doc.rust-lang.org/std/sync/struct.Arc.html#method.into_raw
If we decided to compare addresses by their underlying references, we must also follow the rule:
https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq
```
k1 == k2 -> hash(k1) == hash(k2)
```
Hash is implemented for *const T in the standard library.
Are there any other ways to get a raw pointer hash from a reference?

src/address/channel.rs
```diff
@@ -497,6 +498,16 @@ impl<A: Actor> PartialEq for AddressSender<A> {
 
 impl<A: Actor> Eq for AddressSender<A> { }
 
+impl<A: Actor> Hash for AddressSender<A> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let ptr = Arc::into_raw(Arc::clone(&self.inner));
+        ptr.hash(state);
+
+        // Prevent memory leak.
+        unsafe { Arc::from_raw(ptr); }
```
DoumanAsh: If there is no other way than that please describe in comment why we use this unsafe(while it is obvious better to do it)

fafhrd91:
thanks!

PR #111, opened 2018-07-07
Arbiter: optionally stop system if Actor/Arbiter panics

### LINKED ISSUE #110: Handling panics in Actors run by SyncArbiters
Hi there!

First off, I've been a huge fan of Actix (and Actix Web) so far. Thank you for all of your hard work! :)

Like the issue title says, I'm curious about how to _correctly_ handle panics of Actors running in separate threads.

For some context, I'm writing an IRC relay bot using the irc crate and actix. I have an actor wrapping the future::Stream of events coming from the IRC client. So far so good. The IRC client is running on the main thread sending these events to another actor running on a different thread (via a SyncArbiter):

```rust
let sys = actix::System::new("ircrelay");
let message_handler = Arbiter::start(|_| MessageHandler::with_scrollback_capacity(100));

let message_handler_clone = message_handler.clone();
let cfg_clone = config.clone();
let _irc_actor: Addr<Unsync, IrcActor> =
        actix::Supervisor::start(move |_| IrcActor::new(message_handler_clone, &cfg_clone));
```

From my IrcActor, I send the message to the `MessageHandler`

```rust
if self.should_relay(&irc_message) {
    if let Some(message) = StructuredIrcMessage::from_message(irc_message) {
        self.handler_addr.try_send(message).expect("Whoops, failed to send!");
    }
}
```

At one point I had a panic occur in the recipient as a result of the message, which meant the thread died. It keeps running along despite the panic until another message comes through at which point:

```
thread 'main' panicked at 'Failed to send!: SendError::Closed(..)', libcore/result.rs:945:5
```

Which is expected, but untimely. In terms of wanting to write reliable software, I'd ideally like to recover from that failure as quickly as possible (possibly by killing the process and letting systemd restart the process) rather than having a ticking timebomb waiting for the next message to cause the main thread panic. I'd originally used `do_send` and the process silently failed until I noticed it was failing.

I could wrap my handler bodies in [catch_unwind](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) but that doesn't really appeal to me and if anything that should happen at the framework level as a way of cleanly aborting the process.

What's the right way of dealing with panics in actors in other threads?

Xorlev:
When an Actor panics in an Arbiter, the Arbiter thread stops but no
other action is taken, the rest of the program continues to run. Only if
a sender is using try_send or send will they notice the failure. This
can lead to surprising circumstances where a program will appear to be
running but instead silently misbehaving or ready to explode on the next
message to that Actor (as a panic while processing the sent message does
_not_ fail initially).

Given that a panic is usually indicative of a fairly abnormal condition,
it's not necessarily correct to even restart execution in those cases.
For instance, `Mutex` will poison its self if the MutexGuard is dropped
while the thread is panicking. It's also usually desirable to exit the
program and let process supervision handle reinitializing the program.

This PR adds a new builder API to Arbiter to support the `stop_system_on_panic`
option. Future PRs can also use this builder to allow for further
customization.

Usage:
```rust
    let actor: Addr<SomeActor> = Arbiter::builder()
            .name("optional name")
            .stop_system_on_panic(true)
            .start(|_| SomeActor::new());
```

Or to get an `Addr<Arbiter>`:
```rust
    let arbiter: Addr<Arbiter> = Arbiter::builder()
            .name("optional name")
            .stop_system_on_panic(true)
            .build();
```

The existing API (`Arbiter#new`, `Arbiter#start`) uses the builder under
the hood, but does not change the behavior or API surface of the existing
API.

Resolves #110

### DIFF
--- a/src/arbiter.rs
+++ b/src/arbiter.rs
@@ -30,22 +30,55 @@ thread_local!(
 /// Arbiter controls event loop in it's thread. Each arbiter runs in separate
 /// thread. Arbiter provides several api for event loop access. Each arbiter
 /// can belongs to specific `System` actor.
+///
+/// By default, a panic in an Arbiter does _not_ stop the rest of the System,
+/// unless the panic is in the System actor. Users of Arbiter can opt into
+/// shutting down the system on panic by using `Arbiter::builder()` and enabling
+/// `stop_system_on_panic`.
 pub struct Arbiter {
     stop: Option<Sender<i32>>,
+    stop_system_on_panic: bool,
 }
 
 impl Actor for Arbiter {
     type Context = Context<Self>;
 }
 
+impl Drop for Arbiter {
+    fn drop(&mut self) {
+        if self.stop_system_on_panic && thread::panicking() {
+            eprintln!("Panic in Arbiter thread, shutting down system.");
+            System::current().stop_with_code(1)
+        }
+    }
+}
+
 impl Arbiter {
     /// Spawn new thread and run event loop in spawned thread.
     /// Returns address of newly created arbiter.
+    /// Does not stop the system on panic.
+    pub fn builder() -> Builder {
+        Builder::new()
+    }
+
+    /// Spawn new thread and run event loop in spawned thread.
+    /// Returns address of newly created arbiter.
+    /// Does not stop the system on panic.
     pub fn new<T: Into<String>>(name: T) -> Addr<Arbiter> {
+        Arbiter::new_with_builder(Arbiter::builder().name(name))
+    }
+
+    /// Spawn new thread and run event loop in spawned thread.
+    /// Returns address of newly created arbiter.
+    fn new_with_builder(builder: Builder) -> Addr<Arbiter> {
         let (tx, rx) = std::sync::mpsc::channel();
 
         let id = Uuid::new_v4();
-        let name = format!("arbiter:{}:{}", id.hyphenated().to_string(), name.into());
+        let name = format!(
+            "arbiter:{}:{}",
+            id.hyphenated().to_string(),
+            builder.name.as_ref().unwrap_or(&"actor".into())
+        );
         let sys = System::current();
 
         let _ = thread::Builder::new().name(name.clone()).spawn(move || {
@@ -61,7 +94,10 @@ impl Arbiter {
             // start arbiter
             let addr =
                 rt.block_on(future::lazy(move || {
-                    let addr = Actor::start(Arbiter { stop: Some(stop) });
+                    let addr = Actor::start(Arbiter {
+                        stop: Some(stop),
+                        stop_system_on_panic: builder.stop_system_on_panic,
+                    });
                     Ok::<_, ()>(addr)
                 })).unwrap();
             ADDR.with(|cell| *cell.borrow_mut() = Some(addr.clone()));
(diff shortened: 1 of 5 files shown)


Xorlev:
If this general approach looks good, I'll open another PR updating docs.

src/arbiter.rs
```diff
@@ -97,7 +128,10 @@ impl Arbiter {
 
         // start arbiter
         let ctx = Context::with_receiver(rx);
-        let fut = ctx.into_future(Arbiter { stop: None });
+        let fut = ctx.into_future(Arbiter {
+            stop: None,
+            stop_system_on_panic: true, // If the system Arbiter panics, stop the system.
```
Xorlev: I figured that if this Arbiter failed, it'd be worth stopping. But it'd probably stop anyways with a panic.

src/arbiter.rs
```diff
@@ -217,3 +259,50 @@ impl<I: Send, E: Send> Handler<Execute<I, E>> for Arbiter {
...
+    }
+
+    /// Sets the name of the Arbiter.
+    pub fn name<T: Into<String>>(mut self, name: T) -> Self {
+        self.name = Some(name.into());
+        self
+    }
+
+    /// Sets the option 'stop_system_on_panic' which controls whether the System is stopped when an
+    /// uncaught panic is thrown from an Actor in the Arbiter.
```
DoumanAsh: Add note that by default it is false

src/arbiter.rs
```diff
@@ -32,20 +32,48 @@ thread_local!(
 /// can belongs to specific `System` actor.
```
DoumanAsh: Add few lines about panic behavior for system arbiter and user created ones

DoumanAsh:
I wonder if this behavior may be better as default?

Xorlev:
> I wonder if this behavior may be better as default?

I agree, it's a sane default. Right now, the behavior can be surprising and would require finding this API -- though I agree with the this new behavior being toggleable. @fafhrd91, thoughts?

src/arbiter.rs
Xorlev: Done.

src/arbiter.rs
Xorlev: Done.

fafhrd91:
Let’s disable this by default. We can turn it on in next major release

Xorlev:
@fafhrd91 Done, back to you to merge.

src/arbiter.rs
```diff
@@ -217,3 +264,52 @@ impl<I: Send, E: Send> Handler<Execute<I, E>> for Arbiter {
...
+
+    /// Whether the Arbiter will stop the whole System on uncaught panic. Defaults to false.
+    stop_system_on_panic: bool,
+}
+
+impl Builder {
+    fn new() -> Self {
+        Builder {
+            name: None,
+            stop_system_on_panic: false,
```
DoumanAsh: As I mentioned previously, should we make it true by default?

It is not a problem to change the behavior so suddenly.

fafhrd91: I am not sure. Let’s wait till next major version

DoumanAsh: Ok, then @Xorlev ignore my comment here 😄

fafhrd91:
@Xorlev thanks!

fafhrd91:
why this feature is on? do i need next time to double check all my comment?

PR #113, opened 2018-07-10
Remove obsolete references to the deleted guide

kohensu:
The content of the [guilde](https://actix.rs/book/actix/) is obsolete, moreover, it seems that the guide have been removed from the repository.
So maybe removing the guide references from README and the API doc


fafhrd91:
everything should be up to date

PR #114, opened 2018-07-10
Improve timing related tests

DoumanAsh:
Let's see how appveyor would handle it now


DoumanAsh:
Now tests are not failing, but I wonder if these changed tests are equal to what we had...

PR #119, opened 2018-07-19
run_interval should have initial fire with delay

DoumanAsh:


DoumanAsh:
@fafhrd91 The change here is kinda bug fix for the way interval works(we shouldn't fire callback immediately), do you think it is ok to count it as bugfix?

PR #123, opened 2018-08-01
Add methods to Registry and add back Mocker

kingoflolz:


DoumanAsh:
@kingoflolz Can you please add some tests for methods you added?
Also mention new API in [CHANGES.md](https://github.com/actix/actix/blob/master/CHANGES.md)

fafhrd91:
i have a feeling that set() should be available only during system initialization and not during runtime

DoumanAsh:
@fafhrd91 Hmm well it does make sense... but on other hand `get` also sets actor if it is not present.
One could argue why when we have `query` cannot we let user to `set` by himself

fafhrd91:
well, `query` should be unavailable during runtime too :)

mpapierski:
Hey,

Any update on this? Would splitting it into two separate PRs (1. introduce mocker 2. additional APIs) make it possible to move forward on this?

DoumanAsh:
@fafhrd91 Let's have decision on this PR, should we at least introduce mocker, but leave extra methods out?

mpapierski:
Thanks @DoumanAsh for quick reply. Reintroducing mocker should be nobrainer. I also believe additional methods to query the registry would be quite useful to give user more control over the services.

DoumanAsh:
Yeah I understand, I kinda forgot about this PR so thanks for reminding

mpapierski:
I just noticed that everything is merged. Really cool. Thanks!

PR #127, opened 2018-08-03
Improve documentation of wra_future and ActorFuture

vmalloc:


src/fut/mod.rs
```diff
@@ -44,7 +44,38 @@ use actor::Actor;
 /// Trait for types which are a placeholder of a value that may become
 /// available at some later point in time.
 ///
-/// This is similar to `futures::Future` trait, except it works with `Actor`
+/// `ActorFuture` is very similar to a regular `Future`, except for the fact that `map`ing over it
```
DoumanAsh: > `map`ing over 

I think it is better to re-phrase it like

>combinators accepts closures with third arguments: reference to actor, context and result

Or maybe something along these lines.
I think we should prefer to be generic here rather than using specific combinator `map`

src/fut/mod.rs
```diff
@@ -44,7 +44,38 @@ use actor::Actor;
...
+/// works with closures accepting the relevant actor and its context.
+///
+/// `ActorFuture` is required in cases where future processing requires access to the actor ot its context -
```
DoumanAsh: Rather than saying `required` it is better to say that there are use cases such as: allowing access actor/context from within future, waiting for future within actor's context.

src/fut/mod.rs
```diff
@@ -44,7 +44,38 @@ use actor::Actor;
...
+/// as this cannot be done by capturing a reference as a part of the closure.
+///
+/// Here is an example of a handler on a single actor, deferring work to a different actor, and
```
DoumanAsh: different -> another

src/fut/mod.rs
```diff
@@ -44,7 +44,38 @@ use actor::Actor;
...
+/// impl Handler<DeferredWork> for OriginalActor {
+///     type Result = ResponseActFuture<Self, (), Error>;
+///
+///     fn handle(&mut self, _msg: Refresh, ctx: &mut Context<Self>) -> Self::Result {
+///         let send_to_other = self.other_actor_addr
+///             .send(OtherMessage::new())
+///             .map_err(Error::from);
+///
+///         let update_self =
+///             wrap_future::<_, Self>(send_to_other).map(|result, actor, _ctx| {
```
DoumanAsh: Add note that there is also [into_actor](https://docs.rs/actix/0.7.3/actix/fut/trait.WrapFuture.html#tymethod.into_actor)

src/fut/mod.rs
```diff
@@ -319,7 +356,11 @@ where
     act: PhantomData<A>,
 }
 
-/// Converts normal future into `ActorFuture`
+/// Converts normal future into `ActorFuture`, allowing its processing to
```
DoumanAsh: processing -> combinators

vmalloc:
@DoumanAsh I changed the relevant lines you mentioned and updated the PR. Please let me know if this looks better in your opinion

src/fut/mod.rs
```diff
@@ -44,7 +44,38 @@ use actor::Actor;
 /// Trait for types which are a placeholder of a value that may become
 /// available at some later point in time.
 ///
-/// This is similar to `futures::Future` trait, except it works with `Actor`
+/// `ActorFuture` is very similar to a regular `Future`, only with subsequent combinator closures accepting the relevant actor and its context, in addition to the result.
```
DoumanAsh: Minor thing but I believe `relevant` is not needed here

src/fut/mod.rs
```diff
@@ -44,7 +44,38 @@ use actor::Actor;
...
+///                 // Actor's state updated here
+///                 actor.inner_state.update_from(result);
+///             });
+///         // return the wrapping future
+///         Box::new(returned_future);
+///     }
+/// }
+/// ```
+///
+/// See also [into_actor](trait.WrapFuture.html#tymethod.into_actor), which is similar in functionality
```
DoumanAsh: `which provides future conversion using trait` sounds a bit more correct, as functionality is identical :)

vmalloc:
No problem - fixed and pushed again

DoumanAsh:
thanks

PR #128, opened 2018-08-05
Improve documentation for run_interval

vmalloc:


src/actor.rs
```diff
@@ -383,7 +383,8 @@ where
         self.spawn(TimerFunc::new(dur, f))
     }
 
-    ///Spawns job to execute closure with specified interval
+    /// Spawns a job to execute the given closure periodically, every specified interval
```
DoumanAsh: Hmm I'm not exactly native english speaker, but wouldn't some thing like 

>at fixed interval

Sounds better

instead of

>every specified interval of time

?

vmalloc: Hmmm, yeah. I guess it does. However it should be either "at a fixed interval" or "at fixed intervals". Changing to the former

PR #129, opened 2018-08-09
Do not transfer ownership to the method.

cerceris:
There is no necessity to transfer ownership to `Addr::recipient()`. If we transfer ownership to the method, the following code pattern does not work:
```
let addr = ctx.address();

// Pass Recipient<MessageA>
recipients_a.add(addr.recipient());

// Pass Recipient<MessageB>
recipients_b.add(addr.recipient()); // Compiler error: `addr` value used here after move
```

### DIFF
--- a/src/address/mod.rs
+++ b/src/address/mod.rs
@@ -128,7 +128,7 @@ impl<A: Actor> Addr<A> {
     }
 
     /// Get `Recipient` for specific message type
-    pub fn recipient<M: 'static>(self) -> Recipient<M>
+    pub fn recipient<M: 'static>(&self) -> Recipient<M>
     where
         A: Handler<M>,
         A::Context: ToEnvelope<A, M>,


DoumanAsh:
This changes API, but it should not be breaking change, right @fafhrd91?

fafhrd91:
I am not sure about this one. I remember I explicitly chose to use transfer ownership. In most cases we just convert address to recipient.

fafhrd91:
Maybe context should also implement .recipient()?

cerceris:
> I am not sure about this one. I remember I explicitly chose to use transfer ownership. In most cases we just convert address to recipient.
The ownership transfer does not make sense if we take a look at the method's body. But I agree, it is just a change in favor of convenience.

cerceris:
In general, if we do transfer ownership, we need to create N `Addr` instances to get N `Recipient` instances for a particular address. If we don't, we need to create 1 `Addr` instance to get N `Recipient`s.

fafhrd91:
.recipient() needs to be fixed

DoumanAsh:
@fafhrd91 emmm what do you mean by that? Is current `.recipient()` is incorrect?

fafhrd91:
This is breaking chnage, also decision was explicit about this api

PR #138, opened 2018-09-09
Add test for sending ordered messages to an actor

Firstyear:
Add a test case demonstrating sending messages in order to an actor via a future.

### DIFF
--- a/tests/test_order_msg.rs
+++ b/tests/test_order_msg.rs
@@ -1,0 +1,79 @@
+extern crate actix;
+extern crate futures;
+extern crate tokio;
+
+use actix::prelude::*;
+use futures::Future;
+
+struct Toggle;
+
+impl Message for Toggle {
+    type Result = ();
+}
+
+struct Status;
+
+impl Message for Status {
+    type Result = Result<bool, ()>;
+}
+
+struct MyActor {
+    toggle: bool
+}
+
+impl MyActor {
+    pub fn new() -> Self {
+        MyActor {
+            toggle: false,
+        }
+    }
+}
+
+impl Actor for MyActor {
+    type Context = Context<Self>;
+}
+
+impl Handler<Toggle> for MyActor {
+    type Result = ();
+
+    fn handle(&mut self, _: Toggle, _ctx: &mut Context<MyActor>) {
+        self.toggle = !self.toggle;
+    }
+}
+
+impl Handler<Status> for MyActor {
+    type Result = Result<bool, ()>;
+
+    fn handle(&mut self, _: Status, _ctx: &mut Context<MyActor>) -> Self::Result {
+        Ok(self.toggle)
+    }
+}
+
+
+
+/*
+ * Test sending messages in order to the actor.
+ * This relies on future chaining so that once message A
+ * is complete, we are able to check the content for message B
+ */
+#[test]
+fn test_order_msg() {
+    System::run(|| {
+        let myactor = MyActor::new();
+        let myactor_addr = myactor.start();
+
+        let fut = myactor_addr.send(Toggle)
+            .map_err(|_| ())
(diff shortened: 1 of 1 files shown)


tests/test_order_msg.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+impl Handler<Status> for MyActor {
+    type Result = Result<bool, ()>;
+
+    fn handle(&mut self, _: Status, _ctx: &mut Context<MyActor>) -> Self::Result {
+        Ok(self.toggle)
+    }
+}
+
```
DoumanAsh: Remove extra empty lines(leave only one)

tests/test_order_msg.rs
```diff
@@ -0,0 +1,79 @@
...
+        let myactor = MyActor::new();
+        let myactor_addr = myactor.start();
+
+        let fut = myactor_addr.send(Toggle)
+            .map_err(|_| ())
+            .and_then(move |_res| {
+                myactor_addr.send(Status)
+                    .map_err(|_| ())
+                    .map(|r| {
+                        assert_eq!(r, Ok(true));
```
DoumanAsh: Should the test be more complex?

I think there are several tests that check order of messages already and in most cases they use atomic integers

Firstyear: Which test do you mean? test_sync.rs seems to check the number of messages not the ordering I think. I can't see any example of message ordering being tested. Thanks!

DoumanAsh: Something like https://github.com/actix/actix/blob/master/tests/test_address.rs#L48-L77

What I think is that it would be great if you could illustrate that when you send several messages (e.g. 5) that each of them delivered in order

Firstyear: @DoumanAsh I'm not sure what you mean, do you mean that we validate the order of recieving the messages? The test you have linked only shows that messages were sent in succession, not that ordering was validated - my test validates order because if the toggle message isn't processed first, then the status will fail. Does that help?

DoumanAsh: Ah sorry, what I mean it would be good if you took the test as basis and added validation for order of messages.
Just to illustrate that difference between number of messages wouldn't matter.

After all simply sending two messages is very trivial, on other hand: let's say there would be 15 messages and you could show that each of them are delivered in order would be nice.

Well not necessary to do 15 of course, just 4-5 would be fine 😄 

Though if you have better idea to show more complex example feel free to do it :)

Firstyear: I wish I knew how to write a better/complex example, so far I've found the nested send/map syntax really difficult and hard to work with :( I'd be really appreciative if you had ideas on how to make a more complex test like this work to be honest. 

Well, my test does validate the message order, but I can see what you mean about extending that. I think the code coverage already shows that I'm testing something "extra" but I don't fully understand what either.

DoumanAsh: Hmm to do it simply you could go like that:

1. Extend `Toggle` struct to contain expected bool value, before actually toggling value.
2. Actor receives `Toggle` asserts expected value with current one, then toggles  value
3. Then do it 3 times to just illustrate to reader 😄 

How about it?

Also I think in this case you can create 3 futures and spawn them one by one onto tokio's event loop.
It should simplify workflow as you would't need `and_then`

najamelan: Huh, a boolean seems a poor choice here. How about a sending a uint. The receiver keeps a counter and verifies that `received = last + 1`, then updates last... That being said, normally in an actor model messages arrive in arbitrary order, that's where the name "Arbiter" comes from...

Firstyear: I closed this PR as I won't be merging it ... but thanks for the review anyway

PR #139, opened 2018-09-09
Add shipping example to a single actor with back pressure

botika:


DoumanAsh:
From the example you make, it is a bit unclear what exactly is back pressure here and for what reason...

I think it would be reasonable to illustrate example with relatively big number of messages

botika:
It is the answer to a question in gitter, I asked @fafhrd91 and for this reason I made it. Now update  https://github.com/fafhrd91/fectl/pull/8, close this.

PR #146, opened 2018-09-22
Expose API for custom SyncArbiter

Cocalus:
I'm writing a specialized SyncArbiter. This exposes the private APIs that are needed to move src/sync.rs into a separate crate.

### DIFF
--- a/src/address/mod.rs
+++ b/src/address/mod.rs
@@ -1,7 +1,7 @@
 use std::fmt;
 use std::hash::{Hash, Hasher};
 
-pub(crate) mod channel;
+pub mod channel;
 mod envelope;
 mod message;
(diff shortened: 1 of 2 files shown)


DoumanAsh:
@Cocalus Could you please check if exposed APIs has any docstring. For example `AddressSenderProducer` would be missing it which is not completely nice. I would appreciate if you could add simple documentation for all newly exposed items

Cocalus:
I added some comments then decided to just expose less of the channel API. Looks like rustfmt has changed a bit since some of these were last saved.

DoumanAsh:
Thanks.
for some reason appveyor fails, but it is not related to your PR at all

PR #147, opened 2018-09-28
fix infinite loop in ContextFut::poll() caused by late cancel_future()

yuja:
Canceled handles are generally consumed in the "process items" loop, but
the loop never runs if there's no item to process. As a result, the handles
are left unprocessed and the 'outer loop never resolves. This patch addresses
the issue by discarding the canceled handles if there's no item still alive.

The test case is a bit hairy. It wouldn't finish without the fix. I couldn't
come up with a better idea.

### DIFF
--- a/src/contextimpl.rs
+++ b/src/contextimpl.rs
@@ -376,8 +376,15 @@ where
             }
             self.ctx.parts().handles[1] = SpawnHandle::default();
 
-            // merge returns true if context contains new items
+            // merge returns true if context contains new items or handles to be cancelled
             if self.merge() && !self.ctx.parts().flags.contains(ContextFlags::STOPPING) {
+                // if we have no item to process, cancelled handles wouldn't be
+                // reaped in the above loop. this means self.merge() will never
+                // be false and the poll() never ends. so, discard the handles
+                // as we're sure there are no more items to be cancelled.
+                if self.items.is_empty() {
+                    self.ctx.parts().handles.truncate(2);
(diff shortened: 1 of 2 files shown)


src/contextimpl.rs
```diff
@@ -376,8 +376,15 @@ where
...
 
-            // merge returns true if context contains new items
+            // merge returns true if context contains new items or handles to be cancelled
             if self.merge() && !self.ctx.parts().flags.contains(ContextFlags::STOPPING) {
+                // if we have no item to process, cancelled handles wouldn't be
+                // reaped in the above loop. this means self.merge() will never
+                // be false and the poll() never ends. so, discard the handles
+                // as we're sure there are no more items to be cancelled.
+                if self.items.is_empty() {
+                    self.ctx.parts().handles.truncate(2);
```
DoumanAsh: You need to clarify in comment why it is 2

DoumanAsh:
Thanks for PR! I'll need to check out code myself

src/contextimpl.rs
yuja: Yeah, but maybe by another PR to introduce a constant?

There are a handful uses of the `2`, which is the first index of the handles
to be canceled, IIUC.

DoumanAsh: ah true... everywhere is this magic number 😄 
I guess yeah leave it as it is

fafhrd91:
thanks!

PR #148, opened 2018-09-28
Addr::send2 variant which helps unwrap wrapped results

katyo:
In practice many messages like database queries or network requests has results wrapped using `Result` type (i.e. `Message<Result = Result<ActualResult, QueryError>>`).

In this case we forced to write same code over and over again to get actual result.

For example, we may write something like:

```rust
addr.send(msg)
    .map_err(Either::A) // wrap MailboxError
    // wrap QueryError and unwrap ActualResult
    .and_then(|query_result| query_result.map_err(Either::B))
    .and_then(|actual_result| {
         // doing something here
    })
    .or_else(|wrapped_error| {
        // error handling
    })
```

So we merged both errors using `Either` type from futures.

My solution allows avoid this:

```rust
addr.send2(msg)
    .and_then(|actual_result| {
         // doing something here
    })
    .or_else(|wrapped_error| {
        // error handling
    })
```

Of course, I don't sure that this solution is a right thing.
Of course, we can extend `Future` by using trait to get something like this.

Do you have better ideas?
Better in practical meaning.

### DIFF
--- a/src/address/message.rs
+++ b/src/address/message.rs
@@ -1,6 +1,7 @@
 use std::marker::PhantomData;
 use std::time::{Duration, Instant};
 
+use futures::future::Either;
 use futures::sync::oneshot;
 use futures::{Async, Future, Poll};
 use tokio_timer::Delay;
@@ -32,7 +33,8 @@ where
     M: Message,
 {
     pub(crate) fn new(
-        rx: Option<oneshot::Receiver<M::Result>>, info: Option<(AddressSender<A>, M)>,
+        rx: Option<oneshot::Receiver<M::Result>>,
+        info: Option<(AddressSender<A>, M)>,
     ) -> Request<A, M> {
         Request {
             rx,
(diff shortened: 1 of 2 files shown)


DoumanAsh:
Sorry I kinda forgot about this PR.

Moving on the suggestion, to be honest I'm not sure if it actually improves ergonomics all that much.
Futures provide `flatten` which requires ability for outer error to be convertible to inner error
This method is specifically made to simplify nested futures in such cases, I think it would be better for user to use it, instead of providing another way to send messages.

DoumanAsh:
It depends on inner error type, if it allows conversion from outer to inner error then `flatten()` would work perfectly in this case.

I think the ergonomics of `send` should be improved by means of `futures` rather than on actix side as `flatten()` exists specifically for such issue.

robjtede:
Closing for now since it is quite out of date. If the need still exists we can look at it again.

PR #149, opened 2018-09-29
No longer required to use Syn in README.md example.

ericfindlay:
The example doesn't compile as it is now. I just made the minimum changes to get it running.


fafhrd91:
Thanks

PR #150, opened 2018-10-03
Make SendError Fail when possible.

ethanpailes:
The first step in resolving actix/actix-web#528.

I think there will need to be an actix release between when this PR gets merged and when the other patch goes into actix-web.

### DIFF
--- a/src/address/mod.rs
+++ b/src/address/mod.rs
@@ -1,5 +1,6 @@
 use std::fmt;
 use std::hash::{Hash, Hasher};
+use failure;
 
 pub(crate) mod channel;
 mod envelope;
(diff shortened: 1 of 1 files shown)


DoumanAsh:
Thanks for PR, is this the whole content of PR you'd like to make or are there will be more changes?

ethanpailes:
This is the whole content of the changes I'd like to make to actix. There are more changes that should go into actix-web, but they would not compile until SendError is Fail in actix itself, so they will have to wait until actix gets a new release and actix-web bumps it's actix version.

fafhrd91:
@DoumanAsh could you add changes.md entry

fafhrd91:
Thanks

PR #152, opened 2018-10-05
Fix some clippy warnings

mariocao:
Just an update of the tokio library to the last version, and some minor fixes for some warnings identified by `cargo clippy`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,16 @@
 # CHANGES
 
+## [0.x.x](2018-1x-xx)
+
+- Introduce the `clock` module to allow overriding and mocking the system clock
+  based on `tokio_timer`.
+
+- System now has `System::builder()` which allows overriding the system clock
+  with a custom instance. `Arbiter::builder()` can now also override the system
+  clock. The default is to inherit from the system.
+
+- New utility classes `TimerFunc` and `IntervalFunc` in the `utils` module.
+
 ## [0.7.4] (2018-08-27)
 
 ### Added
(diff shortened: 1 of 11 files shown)


DoumanAsh:
Current actix should use latest tokio already, you're more or less bumping minimum version.

Is it necessary to stop using separate tokio crates and use only meta crate? I don't think there is a need to remove it as we might find it useful later on

mariocao:
You are right. As tokio project moved its crates as sub crates, I thought it could be a good practice to follow less number of versions into a single one. So, if you think it is better like that, close the PR. :smile:

DoumanAsh:
Fixing warnings is still good idea though 👍

mariocao:
I removed the tokio stuff and left only the fixes of clippy warnings :)
I will edit the PR title in order not to be misleading.

src/registry.rs
```diff
@@ -290,10 +290,10 @@ impl SystemRegistry {
     }
 
     /// Add new actor to the registry by address, panic if actor is already running
-    pub fn set<A: SystemService + Actor<Context = Context<A>>>(&self, addr: Addr<A>) {
+    pub fn set<A: SystemService + Actor<Context = Context<A>>>(&self, addr: &Addr<A>) {
```
DoumanAsh: hm... This is technically breaking change as it is public interface.

I think instead we should not clone addr below at line 302.
The effect would be the same

mariocao: :+1:  :)

PR #153, opened 2018-10-05
An implementation of Debug trait for public structs

### LINKED ISSUE #135: Implement Debug for actix structs
Currently it's not possible to autogenerate debug instances for structs containing `Addr` as `Debug` is not implemented for it. For other use-cases it would probably be helpful to implement for other structures as well.

GooseDB:
I've implemented Debug trait for public structures from prelude except for MessageResult and Recipient.
 https://github.com/actix/actix/issues/135

Closes #135

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -18,13 +18,9 @@ where
 
 impl<A: Actor<Context = Context<A>>> fmt::Debug for Context<A> {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        let mb = match self.mb {
-            Some(_) => "Some(_)",
-            None => "None",
-        };
         fmt.debug_struct("Context")
             .field("parts", &self.parts)
-            .field("mb", &format!("{}", mb))
+            .field("mb", &self.mb)
             .finish()
     }
 }
--- a/src/context.rs
+++ b/src/context.rs
@@ -2,6 +2,7 @@ use actor::{Actor, ActorContext, ActorState, AsyncContext, SpawnHandle};
 use address::{Addr, AddressReceiver};
 use arbiter::Arbiter;
 use fut::ActorFuture;
+use std::fmt;
 
 use contextimpl::{AsyncContextParts, ContextFut, ContextParts};
 use mailbox::Mailbox;
@@ -13,6 +14,19 @@ where
 {
     parts: ContextParts<A>,
     mb: Option<Mailbox<A>>,
+}
+
+impl<A: Actor<Context = Context<A>>> fmt::Debug for Context<A> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        let mb = match self.mb {
+            Some(_) => "Some(_)",
+            None => "None",
+        };
+        fmt.debug_struct("Context")
+            .field("parts", &self.parts)
+            .field("mb", &format!("{}", mb))
+            .finish()
+    }
 }
 
 impl<A> ActorContext for Context<A>
--- a/src/contextimpl.rs
+++ b/src/contextimpl.rs
@@ -52,7 +52,7 @@ where
 {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         fmt.debug_struct("ContextParts")
-            .field("flags", &format!("{:?}", self.flags))
+            .field("flags", &self.flags)
             .finish()
     }
 }
--- a/src/contextimpl.rs
+++ b/src/contextimpl.rs
@@ -1,5 +1,6 @@
 use futures::{Async, Future, Poll};
 use smallvec::SmallVec;
+use std::fmt;
 
 use actor::{
     Actor, ActorContext, ActorState, AsyncContext, Running, SpawnHandle, Supervised,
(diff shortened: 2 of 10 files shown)


src/context.rs
```diff
@@ -15,6 +16,19 @@ where
...
 
+impl<A: Actor<Context = Context<A>>> fmt::Debug for Context<A> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        let mb = match self.mb {
+            Some(_) => "Some(_)",
+            None => "None",
+        };
+        fmt.debug_struct("Context")
+            .field("parts", &self.parts)
+            .field("mb", &format!("{}", mb))
```
DoumanAsh: I think `format` is not needed here?

Also shouldn't Mailbox already impl Debug?

src/contextimpl.rs
```diff
@@ -44,6 +45,18 @@ where
...
 }
 
+impl<A> fmt::Debug for ContextParts<A>
+where
+    A: Actor,
+    A::Context: AsyncContext<A>,
+{
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("ContextParts")
+            .field("flags", &format!("{:?}", self.flags))
```
DoumanAsh: Do we need format here?
I think bitflags should impl Debug

src/handler.rs
```diff
@@ -166,6 +167,19 @@ pub struct Response<I, E> {
...
+impl<I, E> fmt::Debug for Response<I, E> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        let item = format!(
+            "{}",
+            match self.item {
+                ResponseTypeItem::Result(_) => "Result(_)",
+                ResponseTypeItem::Fut(_) => "Fut(_)",
+            }
+        );
+        fmt.debug_struct("Response").field("item", &item).finish()
```
DoumanAsh: I think you can put formatting into `match self.item` instead and avoid allocating temporary `item`

src/handler.rs
```diff
@@ -220,6 +234,21 @@ pub struct ActorResponse<A, I, E> {
...
+impl<A, I, E> fmt::Debug for ActorResponse<A, I, E> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        let item = format!(
+            "ActorResponseTypeItem::{}",
+            match self.item {
+                ActorResponseTypeItem::Result(_) => "Result(_)",
+                ActorResponseTypeItem::Fut(_) => "Fut(_)",
+            }
+        );
+        fmt.debug_struct("ActorResponse")
```
DoumanAsh: I think you can put formatting into match self.item instead and avoid allocating temporary item

src/context.rs
GooseDB: I have not found Debug implementation for Mailbox.

DoumanAsh: Oh true, but since it is public API, should we impl Debug for it too?

src/handler.rs
GooseDB: It looks creepy. But if you want to, I can change.

GooseDB:
I don't know how to impl Debug for MB. It has just one parametrized field...''

src/context.rs
GooseDB: I don't know how to impl Debug for MB. It has just one parametrized field...'

DoumanAsh:
@GooseDB I think we can just leave simple text as `Mailbox { capacity: <self.msgs.capacity()>}`

src/handler.rs
DoumanAsh: Not sure I understand, how about to make it like:

```rust

let mut fmt = fmt.debug_struct("Response");

match self.item {
    ResponseTypeItem::Result(_) => fmt.field("item", "Result(_)").finish(),
    ResponseTypeItem::Fut(_) => fmt.field("item", "Fut(_)").finish(),
}
```

I guess there is also problem that debug struct is not the best for printing items like that 😄

GooseDB: Well, there is an allocation of fmt :) Is there a difference?
About the way of printing: why not?)

DoumanAsh: >Well, there is an allocation of fmt :) Is there a difference?

Actually there is because you do double allocation ;)

But I guess let's leave it, no one is going to use `Debug` often

GooseDB:
So, what about this?

src/handler.rs
```diff
@@ -166,6 +167,16 @@ pub struct Response<I, E> {
     item: ResponseTypeItem<I, E>,
 }
 
+impl<I, E> fmt::Debug for Response<I, E> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        let mut fmt = fmt.debug_struct("Response");
+        match self.item {
+                ResponseTypeItem::Result(_) => fmt.field("item", &"Result(_)".to_string()),
```
DoumanAsh: I think `to_string` here is not necessary?

src/handler.rs
```diff
@@ -220,6 +231,16 @@ pub struct ActorResponse<A, I, E> {
     item: ActorResponseTypeItem<A, I, E>,
 }
 
+impl<A, I, E> fmt::Debug for ActorResponse<A, I, E> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        let mut fmt = fmt.debug_struct("ActorResponse");
+        match self.item {
+                ActorResponseTypeItem::Result(_) => fmt.field("item", &"Result(_)".to_string()),
```
DoumanAsh: I think `to_string` here is not necessary?

src/handler.rs
GooseDB: &str has no known size at compile-time.

DoumanAsh: oh damn, I see, it is due to have debug struct helper prints fields... Ok got it

DoumanAsh:
Thank you

fafhrd91:
We need update for changes

DanielJoyce:
Why no debug for Recipient and MessageResult?

PR #166, opened 2018-11-08
Updating trust-dns-resolver dependency to version 0.10.

ajsyp:
trust-dns-resolver 0.10.0 has additional, useful options (e.g., max TTL) that can be passed in via ResolverOpts. These newer options are not available in trust-dns-resolver 0.9.1, which is what Actix currently depends on. This pull request updates Actix to depend on trust-dns-resolver 0.10.

What trust-dns-resolver used to call ResolverFuture is now called AsyncResolver. When creating an AsyncResolver, instead of being given a Future which must be waited on to construct the Resolver, you are immediately given an AsyncResolver, plus a background task that must be spawned to perform DNS lookups (https://docs.rs/trust-dns-resolver/0.10.0/trust_dns_resolver/struct.AsyncResolver.html). This pull request essentially does the renaming of types, plus spawns the background task.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # CHANGES
 
+## [0.7.6] (2018-11-08)
+
+### Changed
+
+- Use `trust-dns-resolver` 0.10.0.
+
 ## [0.7.5] (2018-10-10)
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,8 +61,8 @@ uuid = { version = "0.7", features = ["v4"] }
 tokio-signal = { version = "0.2", optional = true }
 
 # dns resolver
-trust-dns-proto = { version = "^0.4.3", optional = true }
-trust-dns-resolver = { version = "^0.9.1", optional = true }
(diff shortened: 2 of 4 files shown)


ajsyp:
Excellent point @DoumanAsh. Yes, the ResolverOpts struct has new fields, so any other crates which currently try to provide trust-dns-resolver 0.9.1's ResolverOpts to Actix's Resolver will suddenly receive a "wrong struct" problem. Would you like me to bump the patch number to 0.7.6?

DoumanAsh:
@fafhrd91 What do you think? Should we leave it for `0.8`?

fafhrd91:
0.7, I don’t think a lot of people uses custom resolver. Let’s just add entry to migration.md

DoumanAsh:
Ok, @ajsyp Could you please add note on changes about new `trust-dns` in CHANGELOG.md and MIGRATION.md

ajsyp:
@DoumanAsh Done! Let me know if there is anything else you'd like in this PR.

fafhrd91:
thanks!

ajsyp:
No problem! And thanks for all your work in the Actix project!

PR #170, opened 2018-11-17
Echo example

volks73:
As a learning exercise for the Actor model and using actix, I thought I would create a simple "echo" application that echos, forwards, and/or copies the content from STDIN to STDOUT. After completing a working application, I thought it might be helpful for others if it was included as an example for the project.

Since I am still learning, I would appreciate any discussion and feedback on the implementation. Is there (currently) a more idiomatic way of asynchronously reading from STDIN and writing to STDOUT using actix/tokio? I had done some research prior to working on this implementation, and based on this research, I made notes within the module-level comments about reading and writing from standard IO with tokio. I believe based on this research, the implementation in this example is (currently) reasonable, but given the heavy development of the rust asynchronous ecosystem, I am sure I have missed something.

I am also unaware of any existing implementations, or plans, for asynchronously reading from STDIN and writing to STDOUT using actix. There is the [tokio-fs] crate that provides asynchronous file and standard stream adaptation, but there appears to be no actor-based implementation. If acceptable, then maybe the `Stdin` and `Stdout` actors implemented in this example can be moved to the `actors` module of the crate to make it easier for others to include reading and writing to and from standard IO in their actix-based projects? To this end, I have included API-level documentation for the actors within the example.

[tokio-fs]: https://github.com/tokio-rs/tokio/tree/master/tokio-fs

### DIFF
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -14,6 +14,17 @@
 //! example can be stopped by either sending EOF via Ctrl+D or terminating the
 //! process with Ctrl+C.
 //!
+//! To run the example with logging statements for observation of actor
+//! life-cycles, start the example with the following command:
+//!
+//! ```text
+//! $ RUST_LOG=echo=trace cargo run --example echo
+//! ```
+//!
+//! This will print any logging statements within _this_ example. If the
+//! `RUST_LOG=echo=trace` is replaced with `RUST_LOG=trace`, then all logging
+//! statements within all dependencies and _this_ example will be printed.
+//!
 //! # Details
 //!
 //! Three actors are created: (1) [`Stdin`], (2) [`Stdout`], and (3) ['Echo'].
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -56,20 +56,22 @@ use tokio::io;
 
 /// The message sent from the [`Stdin`] actor to a recipient.
 ///
-/// The Stdin actor publishes (sends) this message to a subscriber (recipient)
-/// and it is just a thin wrapper around a series of bytes from STDIN.
+/// The Stdin actor publishes (sends) this message to a subscriber (recipient).
+/// This message is just a thin wrapper around a series of bytes from STDIN.
 ///
 /// [`Stdin`]: struct.Stdin.html
 #[derive(Message)]
 struct Input(pub BytesMut);
 
+// A conversion helper implementation for converting BytesMut into the wrapper
+// Input message.
 impl From<BytesMut> for Input {
     fn from(b: BytesMut) -> Self {
         Input(b)
     }
 }
 
-/// The message received by the [`Stdout`] actor.
+/// The message handled by the [`Stdout`] actor.
 ///
 /// The Stdout actor writes the message, which is just a thin wrapper around a
 /// series of bytes, to STDOUT.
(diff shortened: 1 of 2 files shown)


examples/echo.rs
```diff
@@ -0,0 +1,485 @@
...
+impl<D> Actor for Stdin<D>
+    where D: Decoder<Item=BytesMut, Error=Error> + Send + Clone + 'static
+{
+    type Context = Context<Self>;
+
+    fn started(&mut self, ctx: &mut Self::Context) {
+        trace!("STDIN started");
+        let (tx, rx) = mpsc::unbounded();
+        let codec = self.decoder.clone();
+        thread::spawn(|| {
```
DoumanAsh: I think instead of creating own thread, I think it would be better to use `SyncArbiter`

volks73: Thank you for the suggestion. I started refactoring the example to use the `SyncArbiter`, but I am having some problems. Specifically, the Stdin actor has an `AsyncContext` but it needs a `SyncContext` to work with the SyncArbiter. So, I changed the `Context` type for the Stdin actor and modified the `started` method:

```rust
// ...

impl Actor for Stdin {
    type Context = SyncContext<Self>;

        Self::add_stream(Framed::new(io::stdin(), LinesCodec::default()).map_err(|_| ()), ctx);
    }
}
```

But this yields a similar error for the `SyncContext`. I tried another approach. I created an actor that has a `SyncContext` and converted the `FramedRead` into this actor, leaving the Stdin actor's context as an `AsyncContext`.

```rust
struct StdinStream;

impl Actor for StdinStream {
    type Context = SyncContext<Self>;
}

// ...

impl Actor for Stdin {
    type Context = Context<Self>;

        let stream = SyncArbiter::start(1, || {
            FramedRead::new(io::stdin(), LinesCodec::default())
                 .into_actor(&StdinStream)
        });
        ctx.add_stream(stream);
    }
}
```

I think this is the overall right approach, but I get an error about the `StdinStream` not implementing `actix::Actor`. I cannot figure out how to convert the FramedRead stream into an actor. I removed the generics for the decoder/codec for the time being.

DoumanAsh: Ok, Actually with current approach, it seems not feasible to replace `StreamHandler` at all

examples/echo.rs
```diff
@@ -0,0 +1,485 @@
...
+            info!("Begin STDIN thread");
+            // The `io::stdin()` method, which returns an asynchronous Stdin
```
DoumanAsh: `tokio::io::stdin()` is not usable with the runtime used by actix(current thread)

All read/writes rely on `blocking` calls which require multi threaded runtime

examples/echo.rs
```diff
@@ -0,0 +1,485 @@
...
+            // struct, must be used within a tokio runtime to avoid a
+            // BlockingError.
+            tokio::run(FramedRead::new(io::stdin(), codec)
```
DoumanAsh: Essentially you're making use of three runtimes with, 1 for Stdout, 1 for Stdin and 1 for actix itself.

I don't think it would be good example, instead we should focus on using single runtime of actix.

What you can do is not to use tokio's fs or io with blocking calls, and instead use std's stdin and stdout.
You could similarly read stdin from separate channel and transfer lines using channels

JohnTitor:
I'm going to close this PR since it's quite stale and inactive. But feel free to re-open or submit another PR. Thanks for submitting this!

PR #176, opened 2018-11-25
Add note on API usage

kellytk:


src/address/mod.rs
```diff
@@ -115,8 +115,9 @@ impl<A: Actor> Addr<A> {
     #[inline]
     /// Send asynchronous message and wait for response.
     ///
-    /// Communication channel to the actor is bounded. if returned `Future`
-    /// object get dropped, message cancels.
+    /// Communication channel to the actor is bounded. If returned `Future`
```
DoumanAsh: I'm not sure it is correct statement.
You can use this function correctly regardless of context.

The main point should be is that in `stopping` state your future would get dropped and it should be explicitly be told in Actor lifecycle https://actix.rs/actix/actix/trait.Actor.html#stopping

DoumanAsh:
Should we close it @kellytk ?

PR #179, opened 2018-12-02
Revise Registry comments and examples

NPN:
I removed `MyActor1.start()` from the `SystemRegistry` example, as there would otherwise be two `MyActor1`'s running (since `MyActor2` starts a `MyActor1` when it tries to get `MyActor1` from the registry).

I also fixed some grammar/spelling and removed some extra lines.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -1,8 +1,8 @@
 //! Actors registry
 //!
-//! Actor can register itself as a service. Service can be defined as
-//! `ArbiterService` which is unique per arbiter or `SystemService` which is
-//! unique per system.
+//! An Actor can register itself as a service. A Service can be defined as an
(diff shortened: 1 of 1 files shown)


src/registry.rs
```diff
@@ -166,7 +163,7 @@ impl Registry {
 /// System wide actors registry
 ///
 /// System registry serves same purpose as [Registry](struct.Registry.html),
-/// except it is shared across all arbiters.
+/// except that it is shared across all arbiters.
```
DoumanAsh: Is `that` needed here actually? I thought it is either `except` or `except for`

DoumanAsh: Hm... I'm not sure what is right here because I rarely hear `except that`, regardless thank you

PR #180, opened 2018-12-03
Update crossbeam-channel to 0.3 and parking_lot to 0.7

silwol:
As far as I can see, this should be a minor semver change. I hope I got the migration to the new crossbeam-channel API right so that the new code has the same behavior as the previous version.

### DIFF
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -87,9 +87,8 @@ pub struct SyncArbiter<A>
 where
     A: Actor<Context = SyncContext<A>>,
 {
-    queue: cb_channel::Sender<SyncContextProtocol<A>>,
+    queue: Option<cb_channel::Sender<Envelope<A>>>,
     msgs: AddressReceiver<A>,
-    threads: usize,
 }
 
 impl<A> SyncArbiter<A>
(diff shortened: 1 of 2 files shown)


src/sync.rs
```diff
@@ -224,19 +212,10 @@ where
...
-                    self.state = ActorState::Stopped;
-                    A::stopped(&mut act, self);
-                    return;
-                }
-                Some(SyncContextProtocol::Envelope(mut env)) => {
+                Ok(mut env) => {
                     env.handle(&mut act, self);
                 }
-                None => {
+                Err(_) => {
```
DoumanAsh: Just to be sure, there is no way to detect that error is due to disconnect?

silwol:
I think updating parking-lot consistently with actix-web is a good idea. I have a patch in preparation which also updates the other dependencies there. Will submit that either today or tomorrow.

src/sync.rs
DoumanAsh: I looked myself into it and it seems not, but since the only error case here is likely due to dropping all Senders I don't think it is likely for other error to happen

DoumanAsh:
Thank you

PR #195, opened 2019-01-04
Improve the wording of public API documentation

rotty:
This commit results from proof-reading the API docs throughout the
whole crate, excluding the `fut` module and its children.

Mostly, the changes are in the following areas:

- Use a consistent style, which is also used by the standard
  library, for example:

  - Using just "Registers ..." instead instead of "This method
    registers" for functions.

  - Use punctiation more consistently (no full stops in headings, but
    about everywhere else).

- Missing articles and other minor stylistic issues.

- Some markup issues (e.g. missing backticks for identifiers).

Besides mere wording improvements, which I hope are indeed
improvements (I'm not a native speaker, but dare to say that I'm quite
confident in English), the following tweaks also were made along the
way:

- Export the `Builder` type from the private `arbiter` module, after
  renaming it to `ArbiterBuilder`. Previously, this type was not
  directly visible from outside the crate, and thus also missing from
  `cargo doc` output.

- Adjust a few copy-pasted error strings in `AsyncContext`.

### DIFF
--- a/src/actors/signal.rs
+++ b/src/actors/signal.rs
@@ -1,8 +1,9 @@
-//! An actor implementation of Unix signal handling
+//! An actor implementation of Unix signal handling.
 //!
-//! This module implements asynchronous signal handling for Actix. For each
-//! signal `ProcessSignals` actor sends `Signal` message to all subscriber. To
-//! subscriber, send `Subscribe` message to `ProcessSignals` actor.
+//! This module implements asynchronous signal handling for Actix. For
+//! each signal, the `ProcessSignals` actor sends a `Signal` message
+//! to all subscribers. To subscribe, send a `Subscribe` message to
+//! the `ProcessSignals` actor.
 //!
 //! # Examples
 //!
@@ -70,18 +71,18 @@ use std;
 
 use prelude::*;
 
-/// Different types of process signals
+/// Represents the different types of signals a process can receive.
 #[derive(PartialEq, Clone, Copy, Debug)]
 pub enum SignalType {
-    /// SIGHUP
+    /// `SIGHUP`
     Hup,
-    /// SIGINT
+    /// `SIGINT`
     Int,
-    /// SIGTERM
+    /// `SIGTERM`
     Term,
-    /// SIGQUIT
+    /// `SIGQUIT`
     Quit,
-    /// SIGCHILD
+    /// `SIGCHLD`
     Child,
 }
 
@@ -89,7 +90,7 @@ impl Message for SignalType {
     type Result = ();
 }
 
-/// Process signal message
+/// A message representing a received signal.
 #[derive(Debug)]
 pub struct Signal(pub SignalType);
 
@@ -97,7 +98,7 @@ impl Message for Signal {
     type Result = ();
 }
 
-/// An actor implementation of Unix signal handling
+/// An actor that handles Unix signals.
 pub struct ProcessSignals {
     subscribers: Vec<Recipient<Signal>>,
 }
@@ -191,7 +192,6 @@ impl Message for Subscribe {
     type Result = ();
 }
 
-/// Add subscriber for signals
 impl Handler<Subscribe> for ProcessSignals {
     type Result = ();
 
@@ -200,8 +200,11 @@ impl Handler<Subscribe> for ProcessSignals {
     }
 }
 
-/// Default signals handler. This actor sends `SystemExit` message to `System`
-/// actor for each of `SIGINT`, `SIGTERM`, `SIGQUIT` signals.
+/// Default signals handler.
+///
+/// This actor sends the `SystemExit` message to the `System` actor
+/// for each of the following signals: `SIGINT`, `SIGTERM` and
+/// `SIGQUIT`.
 pub struct DefaultSignalsHandler;
 
 impl Default for DefaultSignalsHandler {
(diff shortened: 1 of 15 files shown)


src/address/mod.rs
```diff
@@ -197,10 +198,12 @@ impl<A: Actor> WeakAddr<A> {
     }
 }
 
-/// `Recipient` type allows to send one specific message to an actor.
+/// The `Recipient` type allows to send one specific message to an
+/// actor.
 ///
-/// You can get recipient with `Addr<_, _>::recipient()` method.
-/// It is possible to use `Clone::clone()` method to get cloned recipient.
+/// You can get a recipient using the `Addr<_, _>::recipient()`
```
DoumanAsh: `Addr<_, _>::` -> `Addr::`

src/address/mod.rs
```diff
@@ -214,30 +217,32 @@ where
...
+    /// Sends a message.
     ///
-    /// Deliver message even if recipient's mailbox is full
+    /// Deliver the message even if the recipient's mailbox is full.
     pub fn do_send(&self, msg: M) -> Result<(), SendError<M>> {
         self.tx.do_send(msg)
     }
 
-    /// Try send message
+    /// Try to send a message.
```
DoumanAsh: Try -> Attempts

src/address/mod.rs
```diff
@@ -214,30 +217,32 @@ where
...
-    /// register current task in receivers queue.
+    /// This method fails if the actor's mailbox is full or
+    /// closed. This method registers the current task in the
+    /// receivers queue.
     pub fn try_send(&self, msg: M) -> Result<(), SendError<M>> {
         self.tx.try_send(msg)
     }
 
-    /// Send message and asynchronously wait for response.
+    /// Send a message and asynchronously wait for a response.
```
DoumanAsh: `Send` -> `Sends`

src/arbiter.rs
```diff
@@ -196,7 +196,7 @@ impl Arbiter {
         });
     }
 
-    /// Executes a future on the current thread.
+    /// Executes a lazily constructed future on the current thread.
```
DoumanAsh: I think it would be more accurate to describe it as:

Executes a function and a future returned by it on the current thread

src/context.rs
```diff
@@ -114,16 +114,17 @@ where
         ContextFut::new(self, act, mb)
     }
 
-    /// Handle of the running future
+    /// Returns a handle of the running future.
```
DoumanAsh: `handle of` -> `handle to` ?

src/handler.rs
```diff
@@ -8,49 +8,49 @@ use arbiter::Arbiter;
 use context::Context;
 use fut::{self, ActorFuture};
 
-/// Message handler
+/// A trait for objects which can handle messages of a specific type.
```
DoumanAsh: object is not common term for Rust

I'd suggest to keep it simple and rephrase it as:

A trait that describes how to handle message

src/handler.rs
```diff
@@ -8,49 +8,49 @@ use arbiter::Arbiter;
...
+    /// The type of value that this handler will return.
     type Result: MessageResponse<Self, M>;
 
-    /// Method is called for every message received by this Actor
+    /// This method is called for every message received by this actor.
     fn handle(&mut self, msg: M, ctx: &mut Self::Context) -> Self::Result;
 }
 
-/// Message type
+/// A trait for objects that represent messages.
```
DoumanAsh: A trait describing message that can be handled by `Actor`

src/handler.rs
rotty: I think my description is ok, see the description of `std::io::Write` for example:

> A trait for objects which are byte-oriented sinks.

I think an "object" in Rust just means an instance of any type.

src/handler.rs
rotty: I've left the "object" in there, but have extended the description a bit, as suggested.

src/actors/signal.rs
```diff
@@ -70,34 +71,34 @@ use std;
...
+/// A message representing a received signal.
 #[derive(Debug)]
 pub struct Signal(pub SignalType);
 
 impl Message for Signal {
     type Result = ();
 }
 
-/// An actor implementation of Unix signal handling
+/// An actor implementing Unix signal handling.
```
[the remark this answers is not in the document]
rotty: Done.

src/actors/signal.rs
```diff
@@ -191,7 +192,7 @@ impl Message for Subscribe {
     type Result = ();
 }
 
-/// Add subscriber for signals
+/// Add subscriber for signals.
```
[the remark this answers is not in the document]
rotty: Done.

src/actors/signal.rs
```diff
@@ -200,8 +201,9 @@ impl Handler<Subscribe> for ProcessSignals {
     }
 }
 
-/// Default signals handler. This actor sends `SystemExit` message to `System`
-/// actor for each of `SIGINT`, `SIGTERM`, `SIGQUIT` signals.
+/// Default signals handler. This actor sends the `SystemExit` message
```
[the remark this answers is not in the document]
rotty: Done.

src/address/mod.rs
```diff
@@ -142,7 +143,7 @@ impl<A: Actor> Addr<A> {
         self.into()
     }
 
-    /// Downgrade to WeakAddr
+    /// Returns a corresponding downgraded `WeakAddr`.
```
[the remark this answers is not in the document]
rotty: Done.

src/address/mod.rs
rotty: Done.

src/address/mod.rs
rotty: Done.

src/arbiter.rs
rotty: I deem "the lazily constructed" wording as accurate, albeit maybe a bit short or even cryptic to some. I've instead expanded on that a bit.

src/context.rs
rotty: Done.

src/address/mod.rs
rotty: Done.

DoumanAsh:
Thanks

PR #199, opened 2019-01-11
Implement MessageResponse for Arc

antoine-de:
This way a message can return an Arc which can be useful.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -1,6 +1,7 @@
 use futures::sync::oneshot::Sender as SyncSender;
 use futures::Future;
 use std::fmt;
+use std::sync::Arc;
 
 use actor::{Actor, AsyncContext};
 use address::Addr;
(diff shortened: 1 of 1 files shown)


DoumanAsh:
Sorry, I forgot about PR.

Can you please clarify what exactly use of `Arc` itself as Message response?
Wouldn't it be rather part of some Future's result or `Result` enum?

antoine-de:
Oups sorry I forgot to give you detail about my problem :confused: 

I'm new with the actor model and with actix so maybe it is indeed a misuse of actix.

The idea is to have an [Actor](https://github.com/antoine-de/transpo-rt/blob/background_reloading/src/dataset_handler_actor.rs#L9) that own an `Arc<Data>` (with `Data` being something I want to avoid to copy).

This Actor can answer 2 messages:
 * `GetData`
 * `UpdateData(data)`

Each actor/worker of my actix-web service can query this actor to have the latest dataset. 
This makes it easy in actix-web to have non blocking async reloading of a working dataset without any mutexes which is quite nice.

For the moment I [return a Result](https://github.com/antoine-de/transpo-rt/blob/background_reloading/src/dataset_handler_actor.rs#L43) but since it is a non failing operation I'd prefer to directly return the `Arc<Data>`.

If it is a problem / misunderstanding on my part, I can continue to use a Result.

Thanks for the great work on actix!

DoumanAsh:
I'm somewhat unsure whether it makes lot of sense to just wrap state in `Arc` and manipulate it through messages.
But I guess there should be nothing preventing us from having it.

@fafhrd91 thoughts?

fafhrd91:
i think Arc is fine

PR #211, opened 2019-02-08
Bump trust-dns-proto to 0.6

fabricedesre:
Currently actix ends up pulling both 0.5 and 0.6:
```
trust-dns-proto v0.5.0
└── actix v0.7.10 (/home/fabrice/dev/actix)

trust-dns-proto v0.6.3
└── trust-dns-resolver v0.10.3
    └── actix v0.7.10 (/home/fabrice/dev/actix)
```
Moving to 0.6 prevents that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "actix"
-version = "0.7.10"
+version = "0.7.11"
 authors = ["Nikolay Kim <fafhrd91@gmail.com>"]
 description = "Actor framework for Rust"
 readme = "README.md"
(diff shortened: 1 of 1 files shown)


DoumanAsh:
Can you please check do we actually use `trust-dns-proto` at all in our code base?
I have suspicion that we do not, but I'm currently at work and cannot it check(will check it myself afterwards if you want)

fabricedesre:
You're right, it's unused. I updated the PR and also bumped the patch version.

DoumanAsh:
Thanks, but I'm not planning to release new version just yet

Issue #52, opened 2018-02-27
Benchmarks?


ssbanerje:
Are there any performance benchmarks for single node message latency/throughput (with varying number of actors) compared to Akka (or other actor frameworks)?

fafhrd91:
there are no any benchmarks that I am aware of. would be interesting to see.
on other hand actix model is different from akka, it is more low level.

ssbanerje:
Ok. What do you mean when you say low level. As I understand it, the overall actix API is similar to https://doc.akka.io/docs/akka/2.5.5/scala/typed.html

fafhrd91:
I am not familiar with akka. if you'd like to make benchmark I can help with actix benchmarks

fafhrd91:
@josephnoir thanks for the links. I am interested in benchmarks as well, but I am busy with web framework at the moment, will add benchmarks a bit later

fafhrd91:
@ehiggs give me couple days, I need to experiment with code

fafhrd91:
@ehiggs what params did you use for test?

fafhrd91:
@ehiggs your code seems right. flow just faster for this type of load.

ehiggs:
@fafhrd91 1000 1000 as in the fdb post.

fafhrd91:
@ehiggs  Yes, figured out. My results similar to yours. Flow just faster.

asafyish:
https://www.techempower.com/benchmarks/

Actix looks pretty good.

ssbanerje:
@ehiggs, @fafhrd91 How many threads do the standard benchmarks use?

As I understand with Actix, the `Arbiter` created inside `let system = System::new("test");` only creates a single thread.

fafhrd91:
@ssbanerje number of threads do not matter, benchmark sends only one message

ssbanerje:
Ok. I was under the impression that each actor injects one message into the ring.

ehiggs:
Hi,
I updated the benchmark code for Actix 0.7.5:

```toml
[package]
name = "actix-bench"
version = "0.1.0"
authors = ["blah@blah.com"]

[dependencies]
actix = "*"
futures = "0.1"
```

```rust
extern crate actix;
extern crate futures;

use actix::*;

use std::time::SystemTime;
use std::env;

/// A payload with a counter
struct Payload(usize);

impl Message for Payload {
    type Result = ();
}

struct Node { 
    limit: usize,
    next: Recipient<Payload>,
}

[...]

PR #213, opened 2019-02-13
Add ring test. Also cargo run --release --example ping doesn't work with LTO

ehiggs:
This should offer a benchmark that can be run against each version and satisfied much of #52.

### DIFF
--- a/examples/ring.rs
+++ b/examples/ring.rs
@@ -43,7 +43,7 @@ impl Handler<Payload> for Node {
     }
 }
 
-fn print_usage_and_exit() {
+fn print_usage_and_exit() -> ! {
     eprintln!("Usage; actix-test <num-nodes> <num-times-message-around-ring>");
     ::std::process::exit(1);
 }
@@ -53,25 +53,21 @@ fn main() {
     if  args.len() < 3 {
         print_usage_and_exit();
     }
-    let mut num_nodes = 2;
-    if let Ok(arg_num_nodes) = args[1].parse::<usize>() {
+    let n_nodes = if let Ok(arg_num_nodes) = args[1].parse::<usize>() {
         if arg_num_nodes <= 1 {
             eprintln!("Number of nodes must be > 1");
             ::std::process::exit(1);
         }
-        num_nodes = arg_num_nodes;
+        arg_num_nodes
     } else {
         print_usage_and_exit();
-    }
-    let num_nodes = num_nodes;
+    };
 
-    let mut ntimes = 1;
-    if let Ok(arg_ntimes) = args[2].parse::<usize>() {
-        ntimes = arg_ntimes;
+    let n_times = if let Ok(arg_ntimes) = args[2].parse::<usize>() {
+        arg_ntimes
     } else {
-        print_usage_and_exit();
-    }
-    let ntimes = ntimes; // make constant;
+        print_usage_and_exit()
+    };
 
     let system = System::new("test");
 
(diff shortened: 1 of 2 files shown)


ehiggs:
I took off lto from profile.release as this breaks examples:

```
$ cargo run --release --example ping
   Compiling actix v0.7.11 (/Users/ehigg90120/src/download/actix)
error: can't perform LTO when compiling incrementally

error: aborting due to previous error

error: Could not compile `actix`.

To learn more, run the command again with --verbose.
```

DoumanAsh:
That's weird, can you remove `profile.release` section completely and see that it works with lto?

opt-level and lto with disabled incremental compilation is anyway default

exgens:
There was an experiment: https://github.com/rust-lang/cargo/pull/6564. Maybe this is the reason?
@ehiggs what version of Rust you using?

DoumanAsh:
Oh damn, suddenly changing defaults... 

@exgens do you know since when it will be included? Is it going to be 1.34?

ehiggs:
```
$ cargo --version
cargo 1.34.0-nightly (245818076 2019-01-27)
```

exgens:
It was just an experiment for a week to collect the data. It's already disabled.
There is a PR for ignore incremental when lto is set https://github.com/rust-lang/cargo/pull/6643 so I think you don't need to worry about it 🙂
But some projects (for example https://github.com/jwilm/alacritty) already disabled incremental for release builds with lto enabled: https://github.com/jwilm/alacritty/commit/7eb0ea82ef491f277b8278b64272dd7f245c5953 🙂

ehiggs:
@exgens 
>It was just an experiment for a week to collect the data. It's already disabled.

Should I remote the lto removal then?

ehiggs:
I've upgraded nightly again and indeed the experiment is over. Sorry about that!

I've pushed another commit that puts LTO back in. I also added a comment explaining what the example does.

The performance is a bit all over the place, though.

```
$ cargo run --release --example ring 1000 1000
    Finished release [optimized] target(s) in 11.32s
     Running `target/release/examples/ring 1000 1000`
Setting up nodes
Reached limit of 1000000 (payload was 1000000)
Time taken: 2.129238
$ cargo run --release --example ring 1000 1000
    Finished release [optimized] target(s) in 0.40s
     Running `target/release/examples/ring 1000 1000`
Setting up nodes
Reached limit of 1000000 (payload was 1000000)
Time taken: 1.987859
$ cargo run --release --example ring 1000 1000
    Finished release [optimized] target(s) in 0.29s
     Running `target/release/examples/ring 1000 1000`
Setting up nodes
Reached limit of 1000000 (payload was 1000000)
Time taken: 2.729968
```

examples/ring.rs
```diff
@@ -0,0 +1,98 @@
...
+            prev_addr = Node{limit: num_nodes * ntimes, next: prev_addr.recipient()}.start();
+        }
+
+        Node{limit: num_nodes * ntimes, next: prev_addr.recipient()}
+    });
+
+    let now = SystemTime::now();
+    system.run();
+    match now.elapsed() {
+        Ok(elapsed) => println!("Time taken: {}.{:06}", elapsed.as_secs(), elapsed.subsec_micros()),
```
DoumanAsh: Could you please add units in this println

examples/ring.rs
```diff
@@ -0,0 +1,98 @@
...
+        if msg.0 >= self.limit {
+            println!("Reached limit of {} (payload was {})", self.limit, msg.0);
+            System::current().stop();
+            return;
+        }
+        self.next.do_send(Payload(msg.0 + 1)).expect("Unable to send payload");
+    }
+}
+
+fn print_usage_and_exit() {
```
DoumanAsh: `print_usage_and_exit() -> !`

examples/ring.rs
```diff
@@ -0,0 +1,98 @@
...
+            ::std::process::exit(1);
+        }
+        num_nodes = arg_num_nodes;
+    } else {
+        print_usage_and_exit();
+    }
+    let num_nodes = num_nodes;
+
+    let mut ntimes = 1;
+    if let Ok(arg_ntimes) = args[2].parse::<usize>() {
```
DoumanAsh: `let n_times = if let Ok(arg_`

examples/ring.rs
```diff
@@ -0,0 +1,98 @@
...
+        ntimes = arg_ntimes;
+    } else {
+        print_usage_and_exit();
+    }
+    let ntimes = ntimes; // make constant;
+
+    let system = System::new("test");
+
+
```
DoumanAsh: Remove extra empty line

examples/ring.rs
```diff
@@ -0,0 +1,98 @@
...
+    println!("Setting up nodes");
+    let _: Addr<_> = Node::create(move|ctx| {
```
DoumanAsh: is `Addr<_>` necessary?

Please remove all explicit types if they are not necessary here and below

DoumanAsh:
Thanks, the thing with performance is quite funny.

I imagine multi-threaded runtime might bring a better result for such test, while with single threaded tokio runtime I'd expect more stable results...

PR #215, opened 2019-02-19
Add connected function to Recipient, Sender and AddressSender

apatrushev:
Recipient abstraction extended to have a way to check aliveness of actual actor.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # CHANGES
 
+## [TBD] (TBD)
+
+### Changed
+
+- Introduced methods `Sender::connected()`, `AddressSender::connected()`
+  and `Recipient::connected()` to check availability of alive actor.
+
 ## [0.7.10] (2019-01-16)
 
 ### Changed
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,15 +1,11 @@
 # CHANGES
 
-## [TBD] (TBD)
+## [0.7.10] (2019-01-16)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


src/address/channel.rs
```diff
@@ -458,6 +460,10 @@ where
     fn hash(&self) -> usize {
         self.inner.as_ref() as *const _ as usize
     }
+
+    fn connected(&self) -> bool {
+        self.connected()
```
DoumanAsh: Hmm just curious is compiler ok with this ambiguous method invocation?
Since both struct and trait has the same method it might cause problems

apatrushev: Some similar methods 10 lines above (`do_send`/`try_send`/`send`) exists here a while.

DoumanAsh:
Thanks, sorry I forgot abut PR.
Can you please update [CHANGES.md](https://github.com/actix/actix/blob/master/CHANGES.md) with mention of added method?

apatrushev:
Done

CHANGES.md
```diff
@@ -1,5 +1,12 @@
 # CHANGES
 
+## [TBD] (TBD)
```
DoumanAsh: Actually next release should be `0.7.10` due to various reasons we didn't published it despite stating concrete date thre.

Could you please move it to 0.7.10, sorry about it

apatrushev: NP. Done.

DoumanAsh:
Thanks

PR #220, opened 2019-03-12
Migrate to actix-rt

fafhrd91:
actix-rt is basically old actix's System and Arbiter but without actors and addresses.

TODO: migration guide

@actix/contributors thoughts


src/handler.rs
```diff
@@ -171,15 +170,16 @@ impl<I, E> fmt::Debug for Response<I, E> {
...
+            ResponseTypeItem::Fut(_) => fmt.field("item", &"Fut(_)".to_string()),
+        }
+        .finish()
     }
 }
 
 impl<I, E> Response<I, E> {
     /// Creates an asynchronous response.
-    pub fn async<T>(fut: T) -> Self
+    pub fn r#async<T>(fut: T) -> Self
```
DoumanAsh: Since it is for new crate version, I think it would be better to remove all `async` keyword usage and just use something like `future`

src/lib.rs
```diff
@@ -91,30 +70,30 @@ pub mod actors;
 pub mod clock;
 pub mod fut;
 pub mod io;
-pub mod msgs;
 pub mod registry;
 pub mod sync;
 pub mod utils;
 
-pub use actor::{
+pub use actix_rt::{Arbiter, System, SystemRunner};
```
DoumanAsh: Should we maybe re-export whole `actix_rt` as `rt` module?

It also has useful `spawn` function

src/registry.rs
```diff
@@ -218,18 +220,21 @@ impl Registry {
...
-    system: Addr<Arbiter>,
-    registry: InnerRegistry,
+    system: Arbiter,
+    registry: HashMap<TypeId, Box<Any + Send>>,
 }
 
-type AnyMapSend = HashMap<TypeId, Box<Any + Send>, BuildHasherDefault<FnvHasher>>;
-type InnerRegistry = Arc<ReentrantMutex<RefCell<AnyMapSend>>>;
+lazy_static::lazy_static! {
+    static ref SREG: ReentrantMutex<RefCell<HashMap<usize, SystemRegistry>>> = {
```
DoumanAsh: Since it is guarded, do we actually need `RefCell`? We could get away without overhead of its runtime borrow check

fafhrd91:
Design questions are out of context for right now. I just want to make it work with ecosystem.

I don’t have much for experiments in this area, but if you could build some prototype we could talk about concrete ideas.

DoumanAsh:
Yeah, I understand. I just gave you idea for future versions

Since with 1.0 actix-web will become independent of `actix` itself I could do some experiments freely.

Current PR is fine as it is

najamelan:
This does not compile right now. Even though the travis results will say this passes on stable, it uses a function that doesn't exist, so it can never compile... so I think the travis configuration needs updating as well.

tests in channel are calling send_fn on Arbiter which doesn't exist...
```rust
error[E0599]: no method named `send_fn` found for type `actix_rt::arbiter::Arbiter` in the current scope
   --> src/address/channel.rs:934:17
    |
934 |             arb.send_fn(move || {
    |                 ^^^^^^^

```

luover:
actix now not 1.0,  Since with 1.0 actix-web will become independent of actix itself,
It will be nice like [riker](https://github.com/riker-rs/riker), only dependent on std::future or futures0.3 ?

src/registry.rs
fafhrd91: we can not this is reentrant mutex, so several owners

src/handler.rs
fafhrd91: good catch

src/registry.rs
DoumanAsh: isn't several owners are possible only within the same thread?

src/lib.rs
fafhrd91: i dont think we need rt

src/registry.rs
fafhrd91: several read-only owners

src/lib.rs
DoumanAsh: ok

src/registry.rs
DoumanAsh: ok

najamelan:
Ok, 
I rebased my work from last night, here is actix without tokio (just keeps tokio-timer). I also removed the resolver...
https://github.com/najamelan/actix/commit/85aae8ba5fc32802e9cf3d002c38a956ee076c54

najamelan:
ps: forgot half of message. Today I will try to make an alternative implementation of actix-rt which compiles to wasm. wasm-bindgen already has an executor, so I hope it will be fast.

src/handler.rs
```diff
@@ -263,7 +266,7 @@ impl<A: Actor, I, E> ActorResponse<A, I, E> {
     }
 
     /// Creates an asynchronous response.
-    pub fn async<T>(fut: T) -> Self
+    pub fn r#async<T>(fut: T) -> Self
```
DoumanAsh: Let's rename it as `fut` as `Response::fut`

src/registry.rs
```diff
@@ -90,6 +86,14 @@ pub struct Registry {
     registry: Rc<RefCell<AnyMap>>,
 }
 
+thread_local! {
+    static AREG: Registry = {
```
DoumanAsh: I wonder if it is ok to have global Registry instead of user owned registry?
Not to mention this registry is per thread so there might be unexpected situations when attempting to read global registry

Do you think original `Registry` wasn't suitable as it is?
I think if user would need global registry he could just use `SystemRegistry`

fafhrd91: Old implement was the same, it uses tls. Also user can not get registry object, only addr of actor via from_registry

DoumanAsh: Oh sorry, ok then

PR #221, opened 2019-03-12
Fix compiler warning because of doc comment in the wrong place

najamelan:
Recently rustdoc started issuing compiler warnings for doc comments above macro expansions. Moving this comment inside the macro resolves that. However, this comment still doesn't show up in the documentation, because the struct in question is not publicly reachable. 

So for consistency I remove the pub. As an alternative, it might be `pub( crate )`. Please modify to your liking.

`cargo test --all-features` runs fine. As some extra test I thought I run all tests on actix-web as well, but I can't get actix-web to compile with a local actix, so ... dunno... 

ps: I'm using nightly compiler

### DIFF
--- a/src/contextimpl.rs
+++ b/src/contextimpl.rs
@@ -10,14 +10,14 @@ use contextitems::ActorWaitItem;
 use fut::ActorFuture;
 use mailbox::Mailbox;
 
-/// internal context state
 bitflags! {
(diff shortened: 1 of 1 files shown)


najamelan:
Note there are a bunch of warnings when running cargo test --all-features on actix-web as well... in nightly

PR #223, opened 2019-03-12
Add derive tests for MessageResponse

samrg472:
Related: https://github.com/actix/actix-derive/pull/14

Awaiting new release of actix-derive before CI tests can pass. I've ran tests using a path dependency to a local clone of actix-derive to make sure the tests pass.


DoumanAsh:
Thanks

PR #226, opened 2019-03-15
Improve SyncArbiter examples and documentation

Firstyear:


src/sync.rs
```diff
@@ -94,8 +99,37 @@ impl<A> SyncArbiter<A>
 where
     A: Actor<Context = SyncContext<A>>,
 {
-    /// Start new sync arbiter with specified number of worker threads.
-    /// Returns address of the started actor.
+    /// Start new `SyncArbiter` with specified number of worker threads.
+    /// Returns the address of the started actor pool.
```
DoumanAsh: >actor pool

From user perspective it is still just single actor though
You cannot really operate the pool itself

src/sync.rs
```diff
@@ -1,71 +1,11 @@
...
-//!         // send 5 messages
-//!         for n in 5..10 {
-//!             addr.do_send(Fibonacci(n));
-//!         }
-//!
-//! #       System::current().stop();
-//!     });
-//! }
-//! ```
+//! Sync Actors are actors that run multiple instances on a thread pool.
```
DoumanAsh: You removed few details about lifetimes of sync actors. Please retain these details

src/sync.rs
```diff
@@ -94,8 +99,37 @@ impl<A> SyncArbiter<A>
 where
     A: Actor<Context = SyncContext<A>>,
 {
-    /// Start new sync arbiter with specified number of worker threads.
-    /// Returns address of the started actor.
+    /// Start new `SyncArbiter` with specified number of worker threads.
+    /// Returns the address of the started actor pool.
+    ///
+    /// ## Example
```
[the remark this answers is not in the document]
Firstyear: When people come into the docs via docs.rs, or are pasting a link, they'll see something like this today:

https://docs.rs/actix/0.7.10/actix/sync/struct.SyncArbiter.html

There *are* no examples, and no guides on how to use the function. When you get a link to start directly:

https://docs.rs/actix/0.7.10/actix/sync/struct.SyncArbiter.html#method.start

Then it's important to have the example right there showing the usage.

I'd rather keep all the examples in here please, especially given that documentation and accessibility seem to be big barriers to entry to the project at the moment.

src/sync.rs
Firstyear: I didn't write this doc, this was me fixing formatting .... I also don't quite understand your comment here.

src/sync.rs
Firstyear: If there are important lifecycle details shouldn't these be mentioned on the start/stop/terminate functions them self then? I think that having them here adds confusion. This is the explanation of the high level concept of what a sync arbiter is, and then if there are lifecycle details, these should be explained on the relevant functions.

Said differently, in most cases does the normal usage case actually care about these lifecycles? IMO, they probably don't ....

src/sync.rs
DoumanAsh: Originally it was
>the started actor.

You changed it to
>started actor pool.

Do not do formatting while chaning stuff then

src/sync.rs
DoumanAsh: Even if most people don't care about lifecycle, it is should remain here for completeness sake.
Do not remove some details just because you think they are not important

src/sync.rs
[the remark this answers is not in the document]
DoumanAsh: Then please make link to sync module instead. You do realize that each example is going through doc test(which slows down `cargo test`).

There is no need to repeat the same example if you could make hyper link like
```
Please refer to [module](../index.html) documentation for example
```

src/sync.rs
Firstyear: I missed that I did the word change at the end there, that's my mistake. I'll fix this.

src/sync.rs
Firstyear: I'm not saying remove it, I'm saying put it in a more relevant section, where it has better context - not to mention it's current wording is confusing and dense. I certainly don't understand what it's trying to tell me :(. 

I think that for a general overview, we shouldn't add too many details, when here, we probably actually want to move the lifecycle part to sync context, which seems like a much more relevant location due to the presence of the stop/terminate/state methods. We could also add a link from SyncArbiter to the SyncContext docs to say "for lifetime of the actors in a SyncArbiter, see ...". I think that would be a better way to structure this documentation.

src/sync.rs
Firstyear: Update: I actually ended up adding a bit extra here to clarify the "single address -> multiple actors" situation, because it could be confusing why there is single actor/address, but sync arbiter has many threads.

src/sync.rs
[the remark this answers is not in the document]
Firstyear: I don't think that this is a justification to remove examples? Saying that they go through tests that causes a slow down seems like we are optimising for a test time, rather than saying "let's give the best documentation possible". It also means people have to play link-diving, and that doesn't aid accessibility of relevant information for users.

I think that examples also have to be focused, and specific to the details we are trying to highlight. Continually pointing at one "mega example" will add confusion, not clarity to the documentation we are trying to show, because people won't know anymore what is or isn't relevant to the goal they are trying to achieve.

I also want to say, these functions may be obvious to you as a library expert, but they are not obvious to others. There are often many questions on gitter and issues raised about confusion on how to operate the library correctly.

src/sync.rs
DoumanAsh: It tells you that you're unable prevent sync actors from being stopped, once it is started (i.e. `stopping` method is called)

You can put it in separate paragraph as a Note

src/sync.rs
```diff
@@ -190,7 +226,32 @@ impl<A> SyncContext<A>
 where
     A: Actor<Context = Self>,
 {
-    /// Create new SyncContext
+    /// Create new SyncContext. This is used insted of impl Actor for your Actor
```
DoumanAsh: I just noticed, `SyncContext::new` is not public, there is no need to add docstring for it

src/sync.rs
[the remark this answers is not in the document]
DoumanAsh: Adding reference on how to use sync actor, should prompt user to click and go read the full example instead of this short one.

I don't understand how asking to click link should make it confusing?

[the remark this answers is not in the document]
Firstyear: Links always confuse people, because you remove context of the documentation from the example which contains a great variety of components. Then it becomes a question of "what is relevant here?".

I think in this case, because it's construction, okay, this example can go. But if there are other cases, I want there to be examples embedded to the docs. For example, the rust docs have examples like https://doc.rust-lang.org/std/result/enum.Result.html#method.or_else on all their methods, rather than referring to a singular example. 

Does that seem like an acceptable approach?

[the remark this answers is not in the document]
DoumanAsh: If method is complex enough, for example it has generic bounds, example would be helpful to clear our it usage.
But for simple methods with clear semantics, I do not see a need for example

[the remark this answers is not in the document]
Firstyear: I think we'll have to "discuss" when we come to these "simple vs complex" methods, because I still think there is a gap in our expectations of simple vs complex (IE I can't work out what a method does from type signatures, I need examples generally, where as others can read type signatures and know exactly what's happening.

So in this case I'll remove this specific example.

PR #228, opened 2019-03-19
Improve example for `ActorFuture`

rivertam:
I felt the example was a little lacking when learning about it, so I think the more explicit comments are helpful here. I also removed a syntax error.

Personally, I think this example is a little too complicated to get the point across. It's somewhat unclear where `ActorFuture` even comes in. The fact that we `send` at all is unnecessary and I would have preferred to use a more generic `Future` to demonstrate that the point of `ActorFuture` is just to add `Actor` data to a `Future` (which has little to nothing to do directly with `send`ing, as far as I can tell).

### DIFF
--- a/src/fut/mod.rs
+++ b/src/fut/mod.rs
@@ -57,21 +57,28 @@ use crate::actor::Actor;
 /// }
 ///
 /// impl Handler<DeferredWork> for OriginalActor {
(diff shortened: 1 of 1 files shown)


rivertam:
Just rebased this. I think it's ready to merge.

PR #231, opened 2019-03-20
Add flush to `FramedWrite` on drop

### LINKED ISSUE #227: No way to flush FramedWrite streams?
I have an actor which reads from TCP using `add_stream` and `StreamHandler`. It also owns a `FramedWrite<WriteHalf<TcpStream>, SomeCodec>`, which it often writes to inside of the `StreamHandler`. Stopping the actor can leave unsent messages in that write stream -- I can understand that simply stopping the actor, it wouldn't necessarily know how to flush the stream. Hence, seems like there should be a flush method available on `FramedWrite`.

I think this relates heavily to #132 which would also help a lot.

EDIT: I'd also expect `.close()` on the framed write to flush, but it doesn't appear to.

Sushisource:
Closes #227

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -391,4 +391,15 @@ impl<T: AsyncWrite, U: Encoder> FramedWrite<T, U> {
     pub fn handle(&self) -> SpawnHandle {
         self.inner.0.borrow().handle
     }
+
+    /// Attempts to write any remaining bytes to the stream and flush it
+    pub fn flush(&self) -> Result<(), std::io::Error> {
+        let mut async_writer = self.inner.1.borrow_mut();
+        let inner = self.inner.0.borrow_mut();
+        if !inner.buffer.is_empty() {
+            async_writer.write(&inner.buffer)?;
+            async_writer.flush()?;
+        }
+        Ok(())
+    }
 }
--- a/src/io.rs
+++ b/src/io.rs
@@ -292,7 +292,8 @@ where
     }
 }
 
-/// A wrapper for the `AsyncWrite` and `Encoder` types.
+/// A wrapper for the `AsyncWrite` and `Encoder` types. The AsyncWrite will be flushed when this
+/// struct is dropped.
 pub struct FramedWrite<T: AsyncWrite, U: Encoder> {
     enc: U,
     inner: UnsafeWriter<T, U::Error>,
@@ -399,7 +400,7 @@ impl<T: AsyncWrite, U: Encoder> Drop for FramedWrite<T, U> {
         let mut async_writer = self.inner.1.borrow_mut();
         let inner = self.inner.0.borrow_mut();
         if !inner.buffer.is_empty() {
-            // Results must be ignored during drop
+            // Results must be ignored during drop, as the errors cannot be handled meaningfully
             let _ = async_writer.write(&inner.buffer);
             let _ = async_writer.flush();
         }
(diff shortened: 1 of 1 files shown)


src/io.rs
```diff
@@ -391,4 +391,15 @@ impl<T: AsyncWrite, U: Encoder> FramedWrite<T, U> {
     pub fn handle(&self) -> SpawnHandle {
         self.inner.0.borrow().handle
     }
+
+    /// Attempts to write any remaining bytes to the stream and flush it
+    pub fn flush(&self) -> Result<(), std::io::Error> {
+        let mut async_writer = self.inner.1.borrow_mut();
```
DoumanAsh: I'm not sure if this would be correct approach.

Right now it is supposed to be written in `WriteFut` which would do both: write remaining data and flush underlying IO.

I think instead of performing flush here, we should notify existing task to force poll.
Note that both `flush` and `write` might return `WouldBlock` in which would be handled in `WriteFut::poll` but not here.

I think the better approach would be to just notify task here and only _flush_ in destructor

Firstyear: Excuse my ignorance here, but having a flush function could be useful for when a full frame is written, and we want to make sure the entire encoded message is sent? I can see for long lived streams, it being useful to be able to flush after a full message payload is written, to ensure that we sent the full message to the other end. How do we currently account for that situation? Or is that handled somewhere in the FrameWrite stream ?

DoumanAsh: Actually I was incorrect a bit, destructor should attempt to write remaining data and flush.

Note that the way writer task is designed, it would continue writing until buffer is clear https://github.com/actix/actix/blob/master/src/io.rs#L174 or any error happens along the way

Then it should flush data.

You cannot wait for task to finish in destructor so yeah, there we can attempt to write all data and flush it, but as for `flush` method we could just force notify current task instead of manually writing data

DoumanAsh: Now I think about it, `flush` method is not necessary because each `write` invocation notifies writer task so it would be flushed along the way.

SO we only need destructor to force write and flush

Firstyear: That seems really nice, because that way the consumer of the api doesn't need to think about flush at all - and we'll only flush once the codec has actually emitted a full frame anyway in that case. So I agree, destructor to force a flush seems correct here.

Sushisource: Makes sense to me. Posted a version implemented in `drop`. I verified this works with my test in my application code, but not clear on how to add a test in actix. Looks like this whole file is largely untested. Happy to add one if you have any input there.

DoumanAsh: Not sure about testing possibility, I'd need to think since we cannot really write to dummy destination

src/io.rs
```diff
@@ -392,3 +392,16 @@ impl<T: AsyncWrite, U: Encoder> FramedWrite<T, U> {
...
     }
 }
+
+impl<T: AsyncWrite, U: Encoder> Drop for FramedWrite<T, U> {
+    fn drop(&mut self) {
+        // Attempts to write any remaining bytes to the stream and flush it
+        let mut async_writer = self.inner.1.borrow_mut();
+        let inner = self.inner.0.borrow_mut();
+        if !inner.buffer.is_empty() {
+            // Results must be ignored during drop
```
DoumanAsh: Comments should explain why the code is done this way, not to just state fact :)

Please state that we cannot possibly handle errors during drop, so if it is not possible to write remaining data, then we drop it.

Sushisource: Changed a bit

DoumanAsh:
Thanks

PR #232, opened 2019-03-24
Add implementations for `Message` for `Arc` and `Box`

rivertam:
I'd like to add tests but there's not really behavior. Happy to add them if you can think of any to run.


fafhrd91:
could update changes.md

rivertam:
Updated

Issue #233, opened 2019-03-24
Add default implementations for Arc and Box


rivertam:
You can't really add this as the user of the library, but we could add it here and I think it would count as expected behavior.

(retroactive issue: one PR already submitted: #232)

DoumanAsh:
Is there anything else to add?

rivertam:
I was thinking it might be reasonable to do an impl for anything that's `Deref<Message>` but I'm not sure if that goes too far.

Firstyear:
Would it be possible to show how to use Box/Arc as Messages in the examples? I think in my mind it's unclear due to the Result type being defined to R, and I don't know how this would look for me to compile/express what the Result type should be here.

DoumanAsh:
@rivertam `Deref<Message>` seems like confusing idea. I'm not sure how you would use it, especially considering that's what is `Box/Arc` is already.

rivertam:
Not sure what you guys are confused about tbh.

@Firstyear It just means `Arc<M>` has the same `Result` as `M`, so sending or handling `Arc<M>` is basically identical to sending or handling `M` itself.

@DoumanAsh I think the `Deref` version would basically just be this

```rust
impl<M, R: 'static> Message for D
where
    M: Message<Result = R>,
    D: std::ops::Deref<Target = M>
{
    type Result = R;
}
```

This could replace the current impls for `Arc` and `Box` because `Arc<M>` and `Box<M>` are already `Deref<Target = M>`. This would just extend that default implementation to anything that derefs to `M`, like `Rc`, `Cow`, etc, but also any user types that deref to `M`.

DoumanAsh:
This is not as simple as just slapping impl for `Deref` and be done with it.
Currently you can have no more than single generic impl, so we should be careful with deciding it.

Not to mention, it might be expected to have such behavior for smart pointers, but not for any other type.
Some people might have own types that actually impl `Deref` but has own result.
So do not just arbitrarily decide for everyone with generic bounds

I'm ok with impls for smart pointers, since they do not carry any meaning outside of owning value, but not generic bounds

rivertam:
Right, so that's why I was saying it might be going too far.

I think it may be fair to say that's essentially what `Deref` means semantically -- it's a wrapper. That said, it doesn't seem quite right for something like `MutexGuard`, which does have a `Deref`. *That* said, if a user *actually* writes out `Handler<MutexGuard<...>>`, I don't really see any harm in letting them do it from just the `Message` perspective.

[...]

PR #234, opened 2019-03-24
Add implementations for `Message` for `Arc` and `Box` (v0.7)

rivertam:
#233


rivertam:
This doesn't need to happen as 0.8 is about to come out. I don't see any harm in it but it probably doesn't matter.

PR #239, opened 2019-04-02
Add Actor::start_in_arbiter.

boxdot:


rivertam:
I'm kind of curious -- shouldn't this take the `Addr` of the arbiter rather than a reference to it directly? When would you have a reference to an `Arbiter` after it's been started?

boxdot:
As far as I can see, in the new implementation of `actix-rt` an arbiter does not have a public address anymore (https://github.com/actix/actix-net/blob/5e8ae210f7e688c065db876565ed7430218785a9/actix-rt/src/arbiter.rs#L42).

rivertam:
@boxdot Got it, thanks. I need to update myself on the new patterns.

PR #244, opened 2019-04-13
Eliminate `extern crate` statements

rotty:
These are no longer needed in the 2018 edition, and their `macro_use`
variants can be replaced with explicitly importing the macros needed.


fafhrd91:
thanks

PR #246, opened 2019-04-13
Add `std::error::Error` implementation for `MailboxError`

rotty:
Since `MailboxError` does not have an underlying error which caused
it, and the `std::error::Error::description()` method is
soft-deprecated ("new `impl`s can omit it", to quote the standard
library), an empty implementation suffices.


fafhrd91:
thanks

PR #247, opened 2019-04-13
examples chat: Add signal handling in server

rotty:
This also serves as an example of how to replace the functionality of
the removed signal actor. An appropriate reference in `CHANGES.md` is
added.


fafhrd91:
thanks

PR #262, opened 2019-06-16
Loosen type bounds on Actor::create

AGausmann:
Currently, `Actor::create` requires that the constructor function satisfies `F: 'static`. 

This was required at first due to the function being called within a spawned future. It first became unnecessary at 12dbad838d92c2a48d01b9cd14192f7f3c887ca0 when `Actor::create` [moved away from that](https://github.com/actix/actix/commit/12dbad838d92c2a48d01b9cd14192f7f3c887ca0#diff-49f917dedc0c847aa6e201bcf53b8143R156), calling the new `Context::create` which [retained the stricter bounds](https://github.com/actix/actix/commit/12dbad838d92c2a48d01b9cd14192f7f3c887ca0#diff-e0934bb6f70a4f8c3baf22a773c14ec7R98) even though the function was now called and dropped within the scope of the call to `create`. 

Not much has changed since then, except in 78455a43d3a7ac696e9e29eb9652002a627d1f62 where `Context::create` was removed and its contents [factored into `Actor::create`](https://github.com/actix/actix/commit/78455a43d3a7ac696e9e29eb9652002a627d1f62#diff-49f917dedc0c847aa6e201bcf53b8143R153).


AGausmann:
More context about my specific use case requiring this: I want to establish a bidirectional channel between two actors,  which I can currently do like this:

```
struct A { b: Addr<B> }
struct B { a: Addr<A> }

impl Actor for A {
    type Context = Context<Self>;
}

impl Actor for B {
    type Context = Context<Self>;
}

fn start() -> Addr<A> {
    A::create(|ctx| {
        let a = ctx.address();
        let b = B::create(move |ctx| {
            B { a }
        });
        A { b }
    });
    a
}
```

However, it is not easy for me (as the third party calling `start()`) to also obtain the address of `B`. With this pull request, I could do the following:

```
fn start() -> (Addr<A>, Addr<B>) {
    let mut b = None;
    A::create(|ctx| {
        let a = ctx.address();
        b = B::create(move |ctx| {
            B { a }
        });
        A { b: b.clone() }
    });
    a
}
```

leizzer:
thank you for contribution and your patience!

Issue #264, opened 2019-06-18
hashbrown 0.5


ignatenkobrain:
actix-web stuff has moved to 0.5, I would appreciate if actix can be moved to 0.5 as well.

Thanks!

JohnTitor:
Close this because `hashbrown` has been deleted from this crate.

PR #265, opened 2019-06-18
chore: Update hashbrown to 0.5

ignatenkobrain:
Closes: #264


fafhrd91:
could you add entry to changes,md

ignatenkobrain:
@fafhrd91 sure, one moment.

ignatenkobrain:
@fafhrd91 done!

JohnTitor:
Looks good to me but the branch is out-of-date.

neoeinstein:
Note, in rustc 1.36, the `std::collections::HashMap` is hashbrown but with the default `SipHash` instead of `FxHash`. We could potentially just use the one in `std` and use `FxHash` as the hasher (or one of our own choosing) where appropriate. Bringing the hasher implementation internal to this crate could then reduce out the additional dependency.

JohnTitor:
Reducing dependencies is good. What about the above comment? @fafhrd91

fafhrd91:
I think this is good idea

robjtede:
probably needs a rebase now the appveyor tests are fixed

JohnTitor:
`hashbrown` has been deleted in #277. If updating `derive_more` is still reasonable, we should drop `hashbrown` from here.

leizzer:
I would like to push this branch to an end. @ignatenkobrain could you update the PR and take @JohnTitor's advice?

PR #266, opened 2019-06-23
Add Context::connected

pandaman64:
If an actor registers an event, the actor does not stop when all addresses to it get dropped.
This behavior is not desirable in some cases like an actor periodically polls an expensive API and works as a cache.
This PR adds `connected` method to `Context`, which allows spawned future to check whether there is an alive address to the actor and to stop it when there is none.


JohnTitor:
Could you also update CHANGES.md?

src/context.rs
```diff
@@ -144,6 +144,10 @@ where
     pub fn set_mailbox_capacity(&mut self, cap: usize) {
         self.parts.set_mailbox_capacity(cap)
     }
+
+    pub fn connected(&self) -> bool {
```
vmalloc: I would add a documentation comment here

pandaman64:
updated.

pandaman64:
Any chance get this merged?

JohnTitor:
I don't have a strong opinion about this, any thoughts? @actix/contributors

PR #274, opened 2019-08-09
Fix few clippy warnings

JohnTitor:
Use `rustfmt` and `clippy` prefixes.


leizzer:
Rerun the build and now it is OK to merge @JohnTitor

PR #277, opened 2019-08-12
Use std::collections::HashMap instead of hashbrown

JohnTitor:
Split from #265 

`std::collections::HashMap`'s implementation has been replaced with `hashbrown::HashMap` implementation in Rust 1.36. Using `std::collections::HashMap` is good because we can reduce dependency.

`hashbrown 0.3.0` uses `fx` algorithm:
> This hash function is very fast for all types of keys, but this algorithm will typically not protect against attacks such as HashDoS.

`std::collections::HashMap` uses `SipHash`:
> While its performance is very competitive for medium sized keys, other hashing algorithms will outperform it for small keys such as integers as well as large keys such as long strings, though those algorithms will typically not protect against attacks such as HashDoS.

I am in favor of using std's default algorithm, but other opinions are welcome.


bitemyapp:
My inclination is to "default safe" unless a key performance need is identified.

Is there an easy way to gather everywhere HashBrown is used so we could get an idea of the current exposure to HashDoS problems?

bitemyapp:
https://github.com/actix/actix/pull/265#issuecomment-520524555 seems to endorse going in this direction as well.

neoeinstein:
Yep, you hit the edit just as I was going to reply. When we're only dealing with `TypeId`s, using `FxHash` is certainly reasonable, but not automatically required.

fafhrd91:
It is fine to use default algorithm for actix, but not for actix-web

JohnTitor:
Okay, this has been approved by some people. I'm going to merge this.

PR #283, opened 2019-09-15
SystemRegistry refactor

chris-ricketts:
Changes to `SystemRegistry`:

- Added `set_checked` function that does not panic if the actor is already present in the registry.
- Removed redundant `get` function.
- Changed `query` to not be a public function as there is currently no way for users to access it.
- Added local `apply_set` function to reduce code repetition.
- Added some integration tests for `set` and `set_checked`

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -273,26 +273,32 @@ impl SystemRegistry {
         }
     }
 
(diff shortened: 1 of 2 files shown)


fafhrd91:
@chris-ricketts do you still want to push this change?

PR #288, opened 2019-09-30
Migrate to std::future

semtexzv:
This contains experimental migration to std::future. In order to get to this, a DNS resolver and ActorRead/Write support has been temporarily disabled. 

In current state, almost all tests pass, and framework is kinda usable. The problem is ActorFuture. I did not implement its combinators, and am currently working on several alternative designs for utilizing `async-await` from within Async actors.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,10 +26,10 @@ path = "src/lib.rs"
 members = ["examples/chat"]
 
 [features]
-default = ["resolver", "http"]
+default = []
 
 # dns resolver
-resolver = ["trust-dns-resolver", "tokio-tcp"]
+resolver = ["trust-dns-resolver", "tokio-net"]
 
 # Adds assertion to prevent processing too many messages on event loop
 mailbox_assert = []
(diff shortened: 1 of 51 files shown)


yawor:
Please correct me if I'm wrong. If I understand this correctly, `ActorFuture` and `ActorStream` were introduced to overcome the borrow checked issue when accessing actor and/or context instances inside the `Future`'s callbacks, as they provide actor and context as arguments to their callbacks.
In case of async/await syntax, there's no need to use callbacks as we can write more linear code so we can access actor and/or context instance directly.
I think what we need to properly utilise async/await is to change `Handler<T>`'s `handle` method to be async. This would allow to await for other futures in `handle`'s code.
Not that I think about it, `ActorFuture` and `ActorStream` may still be needed for some background tasks which can be spawned but don't need to be awaited for.

semtexzv:
The problem is making a implementation that returns futures, which are `'static`. This is required to ensure that the future does not keep references to actors state when it is not running.

The old ActorFuture implementation had this property through using callbacks. After the callback was executed, references were dropped, and everything was OK. 

But, if you try to use current handler implementation with async-await, you will run into a lot of issues with references.

I researched several approaches. One was stateless callbacks chained in async blocks, and several other approaches, but it always resulted in massively reduced readability or performance.

The only approach which seems to be working is rewriting handle method to take `self` by value, consume it on every `await` point, and produce a new one. 

I am working on a green-field implementation, which might be merged back into actix here:
https://github.com/semtexzv/ak

idanarye:
How about something along the lines of:
```rust
fn handle(&mut self, msg: MessageType, ctx: &mut Self::Context) -> AsyncType {
    async_wrapper_glue(async move {
        bla_bla_bla();
        let response = ctx.with_actor(|actor| {
            actor.addr_to_another_actor.send(Some)
        }).await?;
        do_something_with(&response);
        let some_data = ctx.with_actor(|actor| {
            actor.some_hashmap[&response.some_index_field].clone()
        });
        do_stuff_with(some_data);
        Ok(())
    })
}
```

`ctx.with_actor` will be something like:
```rust
pub fn with_actor<T>(dlg: impl FnOnce(&mut A) -> T) { ... }
```
We'll be able to use the actor inside the closure, but we won't be able to escape from it - so if we want to cross `.await` boundaries we'll have to either `clone()` the data from the actor or construct the futures entirely in the closure.

fafhrd91:
hey @semtexzv, just checked your `ak` code, looks super interesting. do you plan to continue?

i'd like to release 0.9 with support of new actix-net and tokio but with ActorFuture. and then work on async/await integration got 1.0 release.

semtexzv:
@fafhrd91 Well, I'd like to, but the approach I've tried is not scalable. The basic problem of allowing mutation of state across await points is still not solved, and If we want rich actors like they are in actix, we need to solve that. 

Alternative course would be to greatly simplify the actors, (Maybe passing the return channel explicitly, not requiring returning value from the handle method). And provide simple api for chaining multiple state mutating method calls.


In the short term, I think moving ActorFuture to `std::future` is a right choice.

flaper87:
> In the short term, I think moving ActorFuture to `std::future` is a right choice.

I'm not the one doing the coding for this but, if I may, I'd like to agree with this statement. I think moving to `std::future` is the right choice. I'm not sure how much work is needed for this but I'd probably do it sooner, assuming the work on `ActorFuture` is not a requirement.

JohnTitor:
This PR is staled, I guess?

frewsxcv:
From what I understand, the migration happened in https://github.com/actix/actix/pull/300, so this can be closed

JohnTitor:
Okay, I'm going to close this PR. But if you have any remaining things, feel free to submit another PR!

PR #303, opened 2019-12-16
Update type signature for ResponseFuture to match BoxFuture

bcmcmill:
The type signature for `ResponseFuture` was causing issues with `Unpin` and `GenFuture`.


fafhrd91:
you should be able to merge yourself

PR #306, opened 2019-12-28
Fix broken Ring example

marhel:
This PR fixes the ring example, which terminates after just a few tens of messages is sent, instead of waiting for all messages to flow through the ring, which makes for a very misleading benchmark (which is why I noticed.)

Also adds some more informative output, such as messages sent per second.

I know very little of actix, as I started playing around with it today, and I expect someone who knows actix to be able to do better, but this is at least an improvement over the current state.

### DIFF
--- a/examples/ring.rs
+++ b/examples/ring.rs
@@ -51,8 +51,9 @@ fn print_usage_and_exit() -> ! {
     ::std::process::exit(1);
 }
 
-#[actix_rt::main]
-async fn main() {
+fn main() -> std::io::Result<()> {
+    let system = System::new("ring");
+
     let args = env::args().collect::<Vec<_>>();
     if args.len() < 3 {
         print_usage_and_exit();
@@ -101,7 +102,9 @@ async fn main() {
             next: prev_addr.recipient(),
         }
     });
-    node.send(Payload(0)).await.unwrap();
+
+    let _req = node.send(Payload(1));
+    system.run();
 
     match now.elapsed() {
         Ok(elapsed) => println!(
@@ -111,4 +114,6 @@ async fn main() {
         ),
         Err(e) => println!("An error occured: {:?}", e),
     }
+
+    Ok(())
 }
--- a/examples/ring.rs
+++ b/examples/ring.rs
@@ -33,9 +33,12 @@ impl Handler<Payload> for Node {
 
     fn handle(&mut self, msg: Payload, _: &mut Context<Self>) {
         if msg.0 >= self.limit {
-            println!("Reached limit of {} (payload was {})", self.limit, msg.0);
+            println!("Actor {} reached limit of {} (payload was {})", self.id, self.limit, msg.0);
             System::current().stop();
             return;
+        }
+        if msg.0 % 498989 == 1 { // prime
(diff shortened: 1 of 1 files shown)


examples/ring.rs
```diff
@@ -47,8 +51,9 @@ fn print_usage_and_exit() -> ! {
     ::std::process::exit(1);
 }
 
-#[actix_rt::main]
-async fn main() {
+fn main() -> std::io::Result<()> {
```
JohnTitor: Hm, this fix needs to remove `async`?

marhel: Well, not really. The way that I got the example to run to completion instead of exiting prematurely was to do system.run() myself, because `node.send(Payload(0)).await.unwrap();` obviously wasn't waiting until all messages had been sent. So I removed `#[actix_rt::main]` to be able to call run().

JohnTitor: Ah, okay, indeed this fix is reasonable, I think.

examples/ring.rs
```diff
@@ -69,38 +74,61 @@ async fn main() {
...
 
-    println!("Setting up nodes");
+    println!("Setting up {} nodes", n_nodes);
+    let limit = n_nodes * n_times;
     let node = Node::create(move |ctx| {
         let first_addr = ctx.address();
         let mut prev_addr = Node {
-            limit: n_nodes * n_times,
+            id: 1,
+            limit: limit,
```
JohnTitor:
```suggestion
            limit,
```

examples/ring.rs
```diff
@@ -69,38 +74,61 @@ async fn main() {
...
             next: first_addr.recipient(),
         }
         .start();
 
-        for _ in 2..n_nodes {
+        for i in 2..n_nodes {
             prev_addr = Node {
-                limit: n_nodes * n_times,
+                id: i,
+                limit: limit,
```
JohnTitor:
```suggestion
                limit,
```

examples/ring.rs
```diff
@@ -69,38 +74,61 @@ async fn main() {
...
         let mut prev_addr = Node {
-            limit: n_nodes * n_times,
+            id: 1,
+            limit: limit,
             next: first_addr.recipient(),
         }
         .start();
 
-        for _ in 2..n_nodes {
+        for i in 2..n_nodes {
```
JohnTitor:
```suggestion
        for id in 2..n_nodes {
```

examples/ring.rs
```diff
@@ -69,38 +74,61 @@ async fn main() {
...
             prev_addr = Node {
-                limit: n_nodes * n_times,
+                id: i,
```
JohnTitor:
```suggestion
                id,
```

examples/ring.rs
```diff
@@ -69,38 +74,61 @@ async fn main() {
...
+                limit: limit,
                 next: prev_addr.recipient(),
             }
             .start();
         }
 
         Node {
-            limit: n_nodes * n_times,
+            id: n_nodes,
+            limit: limit,
```
JohnTitor:
```suggestion
            limit,
```

JohnTitor:
Thanks!

PR #309, opened 2020-01-08
Fix doc example for fut::ActorFuture

Jonathas-Conceicao:
The example for the use of `ResponseActFuture` on `ActorFuture` doc was outdated for actix v0.9, I've swapped the `ignore` for a `norun` on the doc to make the error actually visible by the build system and updated the example so it can work with the most recent version of Actix.

### DIFF
--- a/src/fut/mod.rs
+++ b/src/fut/mod.rs
@@ -48,7 +48,9 @@ use std::task;
 /// Here is an example of a handler on a single actor, deferring work to another actor, and
 /// then updating the initiating actor's state:
 ///
-/// ```rust,ignore
+/// ```rust,no_run
+/// use actix::prelude::*;
+///
 /// // The response type returned by the actor future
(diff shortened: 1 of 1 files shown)


src/fut/mod.rs
```diff
@@ -48,7 +48,9 @@ use std::task;
 /// Here is an example of a handler on a single actor, deferring work to another actor, and
 /// then updating the initiating actor's state:
 ///
-/// ```rust,ignore
+/// ```rust,norun
```
JohnTitor:
```suggestion
/// ```rust,no_run
```

JohnTitor:
Overall looks good except for one nit.

src/fut/mod.rs
Jonathas-Conceicao: Fixed

JohnTitor:
Thanks!

PR #310, opened 2020-01-08
Lower constrain on MessageResponse for ResponseActFuture

Jonathas-Conceicao:
Since Futures doesn't require a Ok and an Err type anymore, I see no reason for the `MessageResponse<A, M>` for a `ResponseActFuture<A, T>` to require that `T` is a `Result`.

The current code also doesn't add any constrains to the `I` and `E` used on the `Result<I, E>`, so using any static type `T` for the `ResponseActFuture` here does seam more appropriated since it's more generic.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -155,11 +155,10 @@ where
     }
 }
 
-impl<A, M, I: 'static, E: 'static> MessageResponse<A, M>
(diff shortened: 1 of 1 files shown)


JohnTitor:
Thanks!

PR #324, opened 2020-02-01
Removed actix-http dependency and http feature

### LINKED ISSUE #321: Decouple actix-http dependency
There is circular dependency: actix -> actix-http -> actix-web -> actix , which makes it hard to update versions. This crate depends on actix-http only because of 2 implementations:

 - https://github.com/actix/actix/blob/ba9f0a374462b80374ab9ecf5956c4a1265e7a87/src/lib.rs#L208
 - https://github.com/actix/actix/blob/ba9f0a374462b80374ab9ecf5956c4a1265e7a87/src/actors/resolver.rs#L151

What about moving those implementations to actix-http crate and remove actix dependency on actix-http breaking the loop.

dunnock:
This fix should resolve #321 to enable moving actix-http specific trait implementations for actors to actix-http crate:

 - https://github.com/actix/actix/blob/ba9f0a374462b80374ab9ecf5956c4a1265e7a87/src/lib.rs#L208
 - https://github.com/actix/actix/blob/ba9f0a374462b80374ab9ecf5956c4a1265e7a87/src/actors/resolver.rs#L151

After this change applied and actix published to crates, we can make actix a dependency for actix-http and implement mentioned traits in the actix-http crate.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,9 @@
 
 ## [NEXT]
 
-CHANGES
+### Changed
 
-* Breaking: removed actix feature `http`. Actix support for http was moved solely to actix-http and actix-web crates.
+* Feature `http` was removed. Actix support for http was moved solely to actix-http and actix-web crates.
 
 ## [0.9.1] 2020-??-??
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # CHANGES
 
+## [NEXT]
+
+CHANGES
+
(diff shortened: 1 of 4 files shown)


JohnTitor:
The build failure is unrelated to this PR but we should fix it. The problem is tokio is optional but isn't marked as `#[cfg(feature = "tokio-compat")]` in 0.18.0-alpha.2 of trust-dns-resolver. I think we can fix it by upgrading the dependency but it has some breaking changes so should be hard. I'll investigate later.

JohnTitor:
It'd be great if you could submit a PR to actix-http!

PR #326, opened 2020-02-05
Improves documentation on how to asynchronously handle messages

Jonathas-Conceicao:
Recently there has been quite a feel people on the repo and on gitter asking and discussing about how to asynchronously handle messages using Actix. This PR mains at improving the documentation of the offered types and implementations and tries to guide uses to the right types upon navigating the docs. I've added some examples to the types in order illustrate a littler better how to use the API as well.

I encourage users to point out any misleading, confusing or incomplete points in the doc related to asynchronous message handling so we can all make the library a littler easier to use.


JohnTitor:
Thanks!

Issue #308, opened 2020-01-08
Use of actor reference on Async Actor handle


Jonathas-Conceicao:
I'm trying implement some asynchronous actor response that uses the Actor's self reference on _actix v0.9_ but I can't get it to work on the account of some lifetime bounds error. I've separated a small example that contains the error:

```Rust
use actix::prelude::*;

#[derive(Message)]
#[rtype(result = "Result<usize, ()>")]
struct Msg;

struct MyActor;

impl Actor for MyActor {
    type Context = Context<Self>;
}

impl Handler<Msg> for MyActor {
    type Result = ResponseActFuture<Self, Result<usize, ()>>;

    fn handle(&mut self, _msg: Msg, _: &mut Context<Self>) -> Self::Result {
        Box::new(self.reply().into_actor(self))
    }
}

impl MyActor {
    async fn reply(&self) -> Result<usize, ()> {
        Ok(42)
    }
}
```
But this gives me the following error:
```
error[E0495]: cannot infer an appropriate lifetime for autoref due to conflicting requirements
  --> src/main.rs:17:23
   |
17 |         Box::new(self.reply().into_actor(self))
   |                       ^^^^^
   |
note: first, the lifetime cannot outlive the anonymous lifetime #1 defined on the method body at 16:5...
  --> src/main.rs:16:5
   |
16 | /     fn handle(&mut self, _msg: Msg, _: &mut Context<Self>) -> Self::Result {
17 | |         Box::new(self.reply().into_actor(self))
18 | |     }
   | |_____^
note: ...so that reference does not outlive borrowed content
  --> src/main.rs:17:18
   |
17 |         Box::new(self.reply().into_actor(self))
   |                  ^^^^
   = note: but, the lifetime must be valid for the static lifetime...
note: ...so that the expression is assignable
  --> src/main.rs:17:9
   |
17 |         Box::new(self.reply().into_actor(self))
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: expected  `std::boxed::Box<(dyn actix::fut::ActorFuture<Actor = MyActor, Output = std::result::Result<usize, ()>> + 'static)>`
              found  `std::boxed::Box<dyn actix::fut::ActorFuture<Actor = MyActor, Output = std::result::Result<usize, ()>>>`
```

I've tried to implement this using the `ActorFuture` and the `them` combinator but I get similar errors, this is what I tried:
```Rust
[...]

PR #334, opened 2020-02-12
PoC of AsyncHandler trait

bekh6ex:
Related to #308 

We discused this problem with @Jonathas-Conceicao in Gitter and I came up with an idea what I would like to have to work easier with async code in actors.

I wanted to tackle this problem: When actor needs to do some async IO, it becomes difficult to manage the state in all the different handlers and code becomes much more complex than needed.

Same result can be achieved with `wrap_future()` and `ctx.wait()` but then the code becomes rather unreadable if there are several async/await blocks involved. This way it's way more straight-forward.

This PR introduces a new trait: `AsyncHandler` with blanket implementation of `Handler` trait.

Idea is that result of `AsyncHandler::handle()` is always a future and actor system would guarantie that while the future is pending, actor wont receive another message, so then actor's state can be safely mutated in the mean time.

Disclaimer: It's Proof-of-concept and literraly my first unsafe code. 😄 

**Note:** Build fails because it runs with all the features enabled and `AsyncHandler` is currently conflicting with `Mocker` giving error:
```
error[E0119]: conflicting implementations of trait `handler::Handler<_>` for type `actors::mocker::Mocker<_>`
```

### DIFF
--- a/src/actors/mod.rs
+++ b/src/actors/mod.rs
@@ -1,5 +1,6 @@
 //! Helper actors
 
+#[cfg(feature = "mocker")]
 pub mod mocker;
 
 #[cfg(feature = "resolver")]
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -3,22 +3,31 @@ use std::future::Future;
 use std::pin::Pin;
 use std::sync::Arc;
 
-use futures::channel::oneshot::{self, Sender as SyncSender};
-use futures::FutureExt;
+use futures::channel::oneshot::{self, Receiver as SyncReceiver, Sender as SyncSender};
+use futures::{FutureExt, future::join3};
 
 use crate::actor::{Actor, AsyncContext};
 use crate::address::Addr;
 use crate::context::Context;
-use crate::fut::{wrap_future, ActorFuture};
+use crate::fut::{ActorFuture, wrap_future};
 use crate::WrapFuture;
 
 pub struct TempRef<T> {
     inner: *mut T,
+    drop_sender: Option<SyncSender<()>>
 }
 
 unsafe impl<T: Send> Send for TempRef<T> {}
 
 impl<T> TempRef<T> {
+    fn new(inner: *mut T) -> (Self, SyncReceiver<()>) {
+        let (s, r) = oneshot::channel::<()>();
+        (TempRef {
+            inner,
+            drop_sender: Some(s),
+        }, r)
+    }
+
     pub fn as_ref(&self) -> &T {
         unsafe {
             let x = self.inner;
@@ -30,6 +39,13 @@ impl<T> TempRef<T> {
             let x = self.inner;
             &mut *x
         }
+    }
+}
+
+impl<T> Drop for TempRef<T> {
+    fn drop(&mut self) {
+        let sender = self.drop_sender.take().unwrap();
+        sender.send(()).unwrap_or_else(|_| panic!("Failed to notify about dropping TempRef. Bug!"))
     }
 }
 
@@ -57,18 +73,20 @@ where
     type Result = ResponseFuture<AsyncResponse<M>>;
 
     fn handle(&mut self, msg: M, ctx: &mut Self::Context) -> Self::Result {
-        let actor = TempRef { inner: self };
+        let (actor, actor_dropped) = TempRef::new(self);
 
-        let tmp_ctx = TempRef { inner: ctx };
+        let (tmp_ctx, ctx_dropped) = TempRef::new(ctx);
 
-        let fut_res = Self::handle(actor, msg, tmp_ctx);
+        let fut_res = async move {
+            Self::handle(actor, msg, tmp_ctx).await
+        };
 
         let (s, r) = oneshot::channel::<M::Result>();
 
         ctx.wait(wrap_future(async move {
-            let res = fut_res.await;
+            let (res, _, _) = join3(fut_res,actor_dropped, ctx_dropped).await;
             s.send(res)
-                .unwrap_or_else(|_| panic!("Failed to send future result"));
+                .unwrap_or_else(|_| panic!("Failed to send future result. Bug!"));
         }));
 
         Box::pin(r.map(|r| AsyncResponse(r.unwrap())))
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -24,8 +24,6 @@ impl<T> TempRef<T> {
             let x = self.inner;
             &*x
         }
-
-        //        todo!()
     }
     pub fn as_mut(&mut self) -> &mut T {
         unsafe {
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -43,19 +43,6 @@ where
 {
     type Result: Future<Output = M::Result> + Sized + 'static;
 
-    //    fn handle(&mut self, msg: (), ctx: &mut Context<MyActor>) {
-    //        use std::sync::Weak;
-    //
-    //        let this = self as *mut Self;
-    //
-    //        let handle = AsyncActorHandle {
-    //            inner: this
-    //        };
-    //
-    //
-    //        Self::async_handle(handle, ());
-    //    }
-
     fn handle(
         actor: TempRef<Self>,
         msg: M,
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -377,7 +377,7 @@ where
     }
 }
 
-pub struct AsyncResponse<M: AsyncMessage>(pub M::Result);
+pub struct AsyncResponse<M: AsyncMessage>(M::Result);
 
 impl<A, M> MessageResponse<A, M> for ResponseFuture<AsyncResponse<M>>
 where
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -4,17 +4,17 @@ use std::pin::Pin;
 use std::sync::Arc;
 
 use futures::channel::oneshot::{self, Receiver as SyncReceiver, Sender as SyncSender};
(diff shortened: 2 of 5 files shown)


src/actors/mod.rs
```diff
@@ -1,5 +1,6 @@
 //! Helper actors
 
+#[cfg(feature = "mocker")]
```
bekh6ex: Had to disable Mocker. Otherwise the compiler complained about conflicting implementations of `Handler`.

src/handler.rs
```diff
@@ -40,6 +104,12 @@ pub trait Message {
     type Result: 'static;
 }
 
+pub trait AsyncMessage
```
bekh6ex: Just a marker. Had to add this trait to not have conflicting implementations.

src/handler.rs
```diff
@@ -307,6 +377,24 @@ where
     }
 }
 
+pub struct AsyncResponse<M: AsyncMessage>(pub M::Result);
```
bekh6ex: Another type to not have conflicting implementations

src/handler.rs
```diff
@@ -3,14 +3,78 @@ use std::future::Future;
...
+    AH: AsyncHandler<M> + 'static,
+    <AH as Actor>::Context: AsyncContext<AH>,
+    M: AsyncMessage<Result = I> + 'static,
+{
+    type Result = ResponseFuture<AsyncResponse<M>>;
+
+    fn handle(&mut self, msg: M, ctx: &mut Self::Context) -> Self::Result {
+        let actor = TempRef { inner: self };
+
+        let tmp_ctx = TempRef { inner: ctx };
```
bekh6ex: As I understand, with actor we can be sure that it is safe to have a mutable reference if we don't call `handle()` on it.

With context I'm not sure. Especially because we are using it below to wait for the future result.

Maybe, wrapper for the context can be something very different, that has limited of different interface, just allowing to get the address, for example.

src/handler.rs
```diff
@@ -3,14 +3,78 @@ use std::future::Future;
...
+
+        let fut_res = Self::handle(actor, msg, tmp_ctx);
+
+        let (s, r) = oneshot::channel::<M::Result>();
+
+        ctx.wait(wrap_future(async move {
+            let res = fut_res.await;
+            s.send(res)
+                .unwrap_or_else(|_| panic!("Failed to send future result"));
+        }));
```
bekh6ex: Here we are blocking processing of messages while future completes.
What also could be done is awaiting for `TempRef` of actor and context to be dropped. This way we can ensure that even if those reverences were sent to some other threads, we still won't have memory issues.

src/handler.rs
```diff
@@ -3,14 +3,78 @@ use std::future::Future;
...
+            &*x
+        }
+    }
+    pub fn as_mut(&mut self) -> &mut T {
+        unsafe {
+            let x = self.inner;
+            &mut *x
+        }
+    }
+}
```
bekh6ex: I'm very unsure about this part, if it is right and correct way to do it.

tests/test_async_actor.rs
```diff
@@ -0,0 +1,67 @@
...
+    ) -> Self::Result {
+        let res = tokio::task::spawn(async move {
+            delay_for(Duration::from_millis(msg.0 as u64)).await;
+            this.as_mut().count += msg.0;
+            this.as_ref().count
+        });
+
+        Box::pin(async move { res.await.unwrap() })
+    }
+}
```
Jonathas-Conceicao: Needing a external dependency (`tokio::task::spawn`)  in order to make use of the new trait doesn't look good.

If ActorFuture was spawned as wait on the context we wold be able to implement this handle as:

```Rust
impl Handler<Num> for MyActor {
    type Result = ResponseActFuture<Self, usize>;

    fn handle(&mut self, msg: Num, ctx: &mut Self::Context) -> Self::Result {
        Box::new(
            (delay_for(Duration::from_millis(msg.0 as u64))
                .into_actor(self)
                .map(move |_, this, _| {
                    this.count += msg.0;
                    this.count
                })),
        )
    }
}
```

More on this in the general notes of the PR.

tests/test_async_actor.rs
```diff
@@ -0,0 +1,67 @@
...
+
+impl AsyncHandler<Num> for MyActor {
+    type Result = Pin<Box<dyn Future<Output = <Num as Message>::Result>>>;
+
+    fn handle(
+        mut this: TempRef<Self>,
+        msg: Num,
+        _ctx: TempRef<Self::Context>,
+    ) -> Self::Result {
+        let res = tokio::task::spawn(async move {
```
Jonathas-Conceicao: If we change this block to:
```Rust
        mut ctx: TempRef<Self::Context>,
    ) -> Self::Result {
        ctx.as_mut().spawn(
            actix::fut::wrap_future(async {})
                .then(move |_, act: &mut Self, _| {
                    act.count += msg.0;
                    delay_for(Duration::from_millis(msg.0 as u64)).into_actor(act)
                })
                .map(move |_, act: &mut Self, _| {
                    act.count -= msg.0;
                }),
        );
        let res = tokio::task::spawn(async move {
```

I think it should be easy to see that the spawned future binded at `res` does not have exclusive access ove actor's internal state and this can lead to unpredictable behavior, so I don't think this implementation correctly follows Actor's semantic.

This happens because the Runtime, not the Actor's Context, is handling the futures that have access over the Actor's self reference and context.

tests/test_async_actor.rs
bekh6ex: There is no need to use an external dependency.

The idea was to show that it can be done. Not always client's async call will be running in the same runtime as actor. (e.g. with async-std one will have to use their runtime ).

tests/test_async_actor.rs
bekh6ex: Ok, I see.
But I mentioned it somehow in comments, that maybe 'context' interface should be minified, in order to avoid such problems. If we expose only 'CTX.address()' method in 'TempRef<Context>' then this problem won't exist.
Personally, I don't see a reason to use 'spawn' method if you can write regular async code. 
If something needs sceduling, then we can have a separate method for this.

bekh6ex: One idea, how this can be solved: with current approach: instead of calling `Self::handle(actor, msg, tmp_ctx);` in sync code we can do this:
```
let fut_res = async move {
    Self::handle(actor, msg, tmp_ctx).await
};

let (s, r) = oneshot::channel::<M::Result>();

ctx.wait(wrap_future(async move {
    let res = fut_res.await;
    s.send(res)
        .unwrap_or_else(|_| panic!("Failed to send future result"));
}));
```
Then we can be sure, that the message processing is already blocked for this actor and `ctx.as_mut().spawn(...);` won't be executed until we finish processing.

src/handler.rs
```diff
@@ -307,6 +377,24 @@ where
...
+    M: AsyncMessage + 'static,
+    M::Result: Send,
+    A::Context: AsyncContext<A>,
+{
+    fn handle<R: ResponseChannel<M>>(self, _: &mut A::Context, tx: Option<R>) {
+        actix_rt::spawn(async move {
+            if let Some(tx) = tx {
+                tx.send(self.await.0)
+            }
+        });
```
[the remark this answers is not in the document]
bekh6ex: I didn't understand the point. When will this code be executed exactly?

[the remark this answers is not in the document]
bekh6ex: Maybe I got it. 
But does it matter in this case if actor's context is already waiting on another future to finish?
So, as I understand, it does not matter where this future is executed, because when the waiting is over we any mutable references anywhere.

tests/test_async_actor.rs
bekh6ex: Implemented in 29296c7. Does it solve it?

bekh6ex:
With last fixes (waiting until the actor reference and context reference are dropped and starting waiting before any of the client code is executed) I don't see a possibility to have more than one mutable references to actor in the same time in the client code, no matter what one does.

But I'm still not sure about the context though.

bekh6ex:
What I meant with previous comment is that there will be 2 mutable references, but I assume `actix` itself does not modify the actor's state, so only one place where it can be modified. So no data race here.

Jonathas-Conceicao:
Actix already supports asynchronous message handling in it's own way. The approach you suggest to it is different from the code base Actix currently have, not exactly easy to maintain, and doesn't bring anymore safety or performance to the current system. I don't think this is the right approach to the problem in hand.

bekh6ex:
Actix has a way to handle asynchronous messages, but not async code.
The reason I started playing with this idea was the cumbersome code I had to write in order to read file contents, modify it and then save.
To do it, I either had to chain `map` calls on futures mixing it with `wrap_future` and `CTX.wait` in the end, or use sync fs API, which I ended up doing in the end to have readable code, but then I have problem with load ballansing between cores.
This way it's relatively easy to use Rust native ` async/await` having API similar to existing one.

Diggsey:
@bekh6ex I created a crate [actix-interop](https://docs.rs/actix-interop/0.1.0/actix_interop/) which has a function `critical_section` which I believe solves your problem very ergonomically.

delef:
https://github.com/rust-lang/rust/pull/115822

Any progress?

robjtede:
The `actix` crate is in maintenance mode, which is to say no one on the team has an interest in developing features for it any more.

I welcome forks of `actix` that try to achieve both speed and ergonomics with an async API, good luck to those motivatied few.

PR #340, opened 2020-02-16
Fix link to the chat example

jduan:
I was on the page https://github.com/actix/actix/tree/master/examples/chat/src wondering what dependencies I should add to 'Cargo.toml`, then I realized that https://github.com/actix/actix/tree/master/examples/chat has everything I need. 

This PR updates the link so it's less confusing.


JohnTitor:
Thanks!

PR #257, opened 2019-05-27
Add address to SyncContext

mmacedoeu:
On scenario where you want to do "recursive" calls this may help

PR #341, opened 2020-02-17
Add address to SyncContext

mmacedoeu:
On scenarios where you want to do "recursive" calls this may help
This is old pull #257 on current master

### DIFF
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -16,8 +16,8 @@ use actix_rt::System;
 use crossbeam_channel as cb_channel;
 use futures::channel::oneshot::Sender as SyncSender;
 use futures::{Future, StreamExt};
-use pin_project::pin_project;
 use log::warn;
+use pin_project::pin_project;
 
 use crate::actor::{Actor, ActorContext, ActorState, Running};
 use crate::address::channel;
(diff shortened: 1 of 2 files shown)


src/sync.rs
```diff
@@ -16,8 +16,8 @@ use actix_rt::System;
 use crossbeam_channel as cb_channel;
 use futures::channel::oneshot::Sender as SyncSender;
 use futures::{Future, StreamExt};
-use pin_project::pin_project;
 use log::warn;
+use pin_project::pin_project;
```
Jonathas-Conceicao: Is pin_project being used here? I can't see it anywhere.

src/sync.rs
```diff
@@ -225,6 +227,7 @@ where
     stopping: bool,
     state: ActorState,
     factory: Arc<dyn Fn() -> A>,
+    address: Addr<A>,
```
Jonathas-Conceicao: I think `crate::address::AddressSenderProducer` would be more suited to be used here instead of `Addr`. The type shows that it doesn't intend to send messages to the Actor itself, but that it can produce address for it. It is used in the [ContextParts](https://github.com/actix/actix/blob/master/src/contextimpl.rs#L42) if you wanna look at some code examples.

src/sync.rs
Jonathas-Conceicao: My bad, I didn't noticed the removed line above. Still think the change should be dropped though, so we can keep git history clean.

src/sync.rs
Jonathas-Conceicao: Yes, the general idea is the same. The `address` method doesn't need to change it's signature at all, just the type internally used by the `SyncContext`. I think it'll make code clearer and similar to how other internal parts of actix works.

Since the SyncContext itself won't be sending message to it's `Actor` it doesn't need to hold an `Addr`. But it does need (as this PR implements) a way to produce new address to it's Actor so whomever access access to the context can also send messages to the Actor, and that's what an `AddressSenderProducer` is for.

Jonathas-Conceicao:
@JohnTitor, as this doesn't introduce any breaking changes, it should be fine to have it merged before the 0.10 gets stabilized, right? Or should we wait for the next release cycle?

JohnTitor:
Also thanks for the review @Jonathas-Conceicao <3

PR #353, opened 2020-02-26
build(deps): use only `futures-{channel,util}` to minimize deps

ErichDonGubler:
Gains from this commit will depend upon other dependencies removing their unnecessary dependencies on `futures` (for instance, doing something similar here for [`actix-rt`](https://github.com/actix/actix-net/pull/104) removes 12 dependencies by itself, but `trust-dns-*` deps only allow 6 of those dep reductions to be realized ATM)). However, they're changes that downstream users can receive without any action necessary from maintainership here, unless there's a breaking change in deps that transitively use `futures`.


ErichDonGubler:
@JohnTitor, I'm not sure why, but I can't add you as a reviewer.

src/actors/resolver.rs
```diff
@@ -3,6 +3,7 @@
 //! ## Example
 //!
 //! ```rust
+//! use futures_util::{future, future::FutureExt};
```
Jonathas-Conceicao: I think this import is not necessary for this example. Probably it's here only due to a rebase as a similar line was been removed recently.

ErichDonGubler: I have removed this. :+1:

ErichDonGubler:
Thanks for the fast turnaround! Wish I was as fast as you. :)

PR #356, opened 2020-03-01
Fix test_restart_sync_actor occasional fail

### LINKED ISSUE #354: test_restart_sync_actor fails randomly
```rust
---- test_restart_sync_actor stdout ----
thread 'test_restart_sync_actor' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`', tests/test_actor.rs:122:5
```

This test failed the same way on i686, aarc64, s390x and ppc64le. Probably there is some race condition, but you better fix it :)

Jonathas-Conceicao:
Test failed due to race condition where the test would finish before the actor had a change to stop. This moves the stop call to the actor, so it's always
called only when the actor actually stops.

One problem with this is that if something breaks on the lib cause the `Actor` to not stop as they should the test would not stop. @JohnTitor would this be a problem on the CI? I'm not sure what problems this would cause. If this is a problem I can add the code bellow to the test so least it's would be easy to see if tests start to break due to timeout, rather than just failing at the stopping count:
```Rust
        actix_rt::spawn(async move {
            delay_for(Duration::from_secs(5)).await;
            panic!("System has not stopped time, are the actors stopping?")
        });
```

fixes #354

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -57,7 +57,6 @@ jobs:
         with:
           command: test
           args: --all --all-features --no-fail-fast -- --nocapture
-                --skip=test_restart_sync_actor
 
       - name: Generate coverage file
         if: matrix.version == 'stable' && github.ref == 'refs/heads/master'
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -60,7 +60,6 @@ jobs:
         with:
           command: test
           args: --all --all-features --no-fail-fast -- --nocapture
-                --skip=test_restart_sync_actor
                 --skip=test_start_actor_message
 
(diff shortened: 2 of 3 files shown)


Jonathas-Conceicao:
@ignatenkobrain would you mind testing this too to see if test is stable now?

JohnTitor:
I'll take a look, could you remove args for skipping from all workflows?

https://github.com/actix/actix/blob/0765e56393d26af48fe78b8993a347a30dae336a/.github/workflows/linux.yml#L60

Jonathas-Conceicao:
Removed it from linux and windows workflows, let me know if I missed any other.

ignatenkobrain:
This fixes #354 for me, thanks!

PR #357, opened 2020-03-01
Add new AtomicResponse

Jonathas-Conceicao:
It has been pointed out and discussed on #308, #334 and on gitter a possible data race when handling messages asynchronously and a note about it has been added to the docs on #326. This offers a easy to use `MessageResponse` to the users that will be spawned into the Actor's context with a `AsyncContext::wait` call.

`AtomicResponse` allows messages responses generated by an `Actor` to have exclusive access over it, making it easier to process messages that need to stop the Actor from processing other futures but still leaving the runtime free to `poll` other spawned futures.

I also removed some unnecessary future wrapping used on other `MessageResponse` implementations.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -9,7 +9,6 @@ use crate::actor::{Actor, AsyncContext};
 use crate::address::Addr;
 use crate::context::Context;
 use crate::fut::ActorFuture;
-use crate::WrapFuture;
 
 /// Describes how to handle messages of a specific type.
 ///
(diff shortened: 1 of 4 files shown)


src/handler.rs
```diff
@@ -84,9 +83,85 @@ where
...
+/// #[derive(Message)]
+/// #[rtype(result = "usize")]
+/// struct Msg;
+///
+/// struct MyActor(usize);
+///
+/// impl Actor for MyActor {
+///     type Context = Context<Self>;
+/// }
+///
```
JohnTitor: Could you hide them? I think it's a bit long to be displayed on the docs.

PR #366, opened 2020-03-15
Use `once_cell` over `lazy_static`

JohnTitor:
Use `once_cell` over `lazy_static` to acquire macro-free. This adds additional dependency but I'm fine with that since the community seems to prefer `once_cell` nowadays.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/actix/actix.git"
 documentation = "https://docs.rs/actix/"
 categories = ["network-programming", "asynchronous"]
 license = "MIT/Apache-2.0"
-exclude = [".gitignore", ".cargo/config"]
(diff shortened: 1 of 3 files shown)


cetra3:
I don't think we support actix in `no-std` do we? Otherwise no issues from me.

JohnTitor:
> I don't think we support actix in no-std do we?

Yeah but I think `once_cell` is more flexible than `lazy_static` and maybe in std in the future so I would like to use and test on actix project.

If someone has any concerns (e.g. dependency addiction), feel free to point out.

Jonathas-Conceicao:
I personally prefer once_cell over lazy_static and I have been moving my  projects to it. I'm fine with moving actix to it too.

JohnTitor:
Okay, seems there is no concern, I'll merge this. Thanks all!

PR #374, opened 2020-04-04
Only poll dropped ContextFut if event loop is running

### LINKED ISSUE #372: May panic in finalization when the context is used in `Actor::started()`
## Version

* v0.9.0
* master

## Platform

This issue occurs at least on:

* Linux
* macOS

## Description

In some situations, `Actor::started()` may be called in the finalization.  In these cases, using the context in `Actor::started()` panics.
https://github.com/masnagam/rust-case-studies/blob/master/actix-started-in-drop/src/main.rs

## Expected Behavior

`Actor::started()` shouldn't be called in the finalization.

## Workaround

https://github.com/masnagam/rust-case-studies/tree/master/actix-started-in-drop#workaround

Jonathas-Conceicao:
This ensures the `ContextFut` will not run when dropped if the event loop has already stopped. Which could cause crash in case of the runtime stopping with a spawned actor that didn't fully start.

fixes #372

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,10 @@
 
 * Update `tokio-util` dependency to 0.3, `FramedWrite` trait bound is changed. [#365]
 
+* Update `actix-rt` dependency to 1.1. [#374]
+
 [#365]: https://github.com/actix/actix/pull/365
+[#374]: https://github.com/actix/actix/pull/374
 
 ## Fixed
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -6,7 +6,10 @@
 
 * Update `tokio-util` dependency to 0.3, `FramedWrite` trait bound is changed. [#365]
 
+* Update `actix-rt` dependency to 1.1. [#374]
```
JohnTitor, line 9: The change looks good! I think it doesn't explain what the bug is, the PR title is more suitable.

JohnTitor:
```suggestion
* Only poll dropped ContextFut if event loop is running. [#374]
```

PR #376, opened 2020-04-06
Fix ActorFuture::poll_next impl for StreamThen

### LINKED ISSUE #375: StreamThen combinator for ActorStream is broken
The `take` on the first line here:

https://github.com/actix/actix/blob/b7c7237223dae46d27426799a48b4354c4376f54/src/fut/stream_then.rs#L67-L73

Means that if the future doesn't complete immediately on first poll it will just be dropped. This is obviously incorrect behaviour: the future shouldn't be removed from the option until it return `Ready` from `poll`.

Jonathas-Conceicao:
Old implementation would lose the stored future if it wasn't immediately ready
as reported on #375

I've also updated the `test_stream_timeout` to also use the `then` method so we ensure it doesn't break again in the future.

fixes #375

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,12 @@
 
 [#365]: https://github.com/actix/actix/pull/365
 
+## Fixed
+
+* Fix `ActorFuture::poll_next` impl for `StreamThen` to not lose inner future when it's pending. [#376]
+
+[#376]: https://github.com/actix/actix/pull/376
+
 ## [0.10.0-alpha.2] 2020-03-05
 
 ## Added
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -8,6 +8,12 @@
 
 [#365]: https://github.com/actix/actix/pull/365
 
+## Fixed
+
+* Fix ActorFuture::poll_next impl for StreamThen to not lose inner future when it's pending. [#375]
+
+[#375]: https://github.com/actix/actix/pull/375
```
JohnTitor, line 15:
```suggestion
* Fix ActorFuture::poll_next impl for StreamThen to not lose inner future when it's pending. [#376]

[#376]: https://github.com/actix/actix/pull/376
```

PR #382, opened 2020-05-09
Make CI green even if coverage gets worse

JohnTitor:
ref. https://docs.codecov.io/docs/commit-status


JohnTitor:
I set `threshold: 100%` at this time but we could tweak this value to a more appropriate one later.

PR #384, opened 2020-05-11
buffer write calls in SinkWrite

robjtede:
fix actix/actix-web#1479

Moves `Sink::start_send` calls into the ActorFuture impl, and correctly guards with `Sink::poll_ready`, buffering items to be sent.

I think this is the key _(from `futures::sink::Sink::start_send` docs)_:

> Each call to this function must be preceded by a successful call to poll_ready which returned Poll::Ready(Ok(())).

Seems that the easiest way to achieve this is as implemented here; with an item buffer and calling poll_ready in the place where task context is available.

The existing `test_send_2` test seems to adequately test this implementation.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 ## Changed
 
+* `SinkWrite::write` calls now send all items correctly using an internal buffer. [#384]
+
 * Update `tokio-util` dependency to 0.3, `FramedWrite` trait bound is changed. [#365]
 
 * Only poll dropped ContextFut if event loop is running. [#374]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 
 ## Changed
 
-* `SinkWrite::write` calls now send all items correctly using an internal buffer. [#384]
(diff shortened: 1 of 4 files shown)


robjtede:
It seems like the only reason that tests in test_sink were working at all is because they were doing this kind of buffering on top.

The custom Sink in those tests does also call `.wake_by_ref` on the current waker when the queue isn’t empty but I’m not convinced it’s necessary.

Jonathas-Conceicao:
>The custom Sink in those tests does also call .wake_by_ref on the current waker when the queue isn’t empty but I’m not convinced it’s necessary.

I think the wake call is still necessary since it will be returning `Poll::Pending`, otherwise the future would not be polled.

robjtede:
> I think the wake call is still necessary since it will be returning Poll::Pending, otherwise the future would not be polled.

@Jonathas-Conceicao Ahh because in this test case, there isn't anything like mio waking up the waker with network activity...?

Jonathas-Conceicao:
That is my understanding.

PR #385, opened 2020-05-11
add Clone trait to WeakAddr

zyctree:


robjtede:
can these structs not be `#[derive(Clone)]` ?

zyctree:
if `#[derive(Clone)]`, `A` need to be  `Clone`, see https://github.com/rust-lang/rust/issues/41481

robjtede:
Right, of course. Thanks for the contribution.

PR #394, opened 2020-05-27
Update `actions/cache` to v2

JohnTitor:
Also shrink `~.cargo/` items.


JohnTitor:
Er, we should have used `actions-rs/cargo` in installing cargo-cache...

JohnTitor:
Added an old workaround, I thought this was unnecessary with 1.40.0 or higher but CI failed unexpectedly, hmm perhaps we should tweak pre-installed rustup/cargo env? Anyways this should work fine now.

PR #400, opened 2020-06-30
Add GroupWeakAddr.

alvardes:
This is useful for when you might have some group of "sibilings" of "child" actors that you want to notify, but don't want to extend their lifetime by holding an `Addr` of them.

Example could be when you have a `Listener`/`Server` which for a particular message/event creates a "child" actor.

The "child" actors could have a shorter lifespan than the `Listener` but they shouldn't outlive the `Listener`. In that case the `Listener` might want to store `WeakAddr`s for all of the "child" actors created and upon stopping it will send a message to all still existing "child" to notify them that they must shut down.

The reason for the pull request is that for this purpose you cannot use a simple `HashSet<WeakAddr<A>>` because `WeakAddr` doesn't (and cannot easily) implement `Hash` and `Eq`.

This PR's `GroupWeakAddr` internally uses the `PtrWeakHashSet` from the external crate `weak_table`.

All reviews are welcome. I am particularly uncertain about what is the best name for this struct so all suggestions are appreciated.

### DIFF
--- a/src/address/group.rs
+++ b/src/address/group.rs
@@ -1,0 +1,214 @@
+use std::hash::Hash;
+use std::ops::Deref;
+
+use crate::actor::Actor;
+use crate::address::envelope::ToEnvelope;
+use crate::address::{Addr, SendError, WeakAddr};
+use crate::handler::{Handler, Message};
+
+use weak_table::{traits::WeakElement, PtrWeakHashSet};
+
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+struct WeakTableAddrWrap<A: Actor>(Addr<A>);
+
+impl<A: Actor> Deref for WeakTableAddrWrap<A> {
+    type Target = Addr<A>;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+#[derive(Debug)]
+struct WeakTableWeakAddrWrap<A: Actor>(WeakAddr<A>);
+
+impl<A: Actor> Deref for WeakTableWeakAddrWrap<A> {
+    type Target = WeakAddr<A>;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl<A: Actor> WeakElement for WeakTableWeakAddrWrap<A> {
+    type Strong = WeakTableAddrWrap<A>;
+
+    fn new(view: &Self::Strong) -> Self {
+        WeakTableWeakAddrWrap(view.0.downgrade())
+    }
+
+    fn view(&self) -> Option<Self::Strong> {
+        self.upgrade().map(|addr| WeakTableAddrWrap(addr))
+    }
+}
+
+impl<A: Actor> Clone for WeakTableWeakAddrWrap<A> {
+    fn clone(&self) -> WeakTableWeakAddrWrap<A> {
+        WeakTableWeakAddrWrap(self.0.clone())
+    }
+}
+
+/// A group of actors that doesn't extend the actors lifetime. This uses `WeakAddrs` inside.
+///
+/// You can use this to send messages to a group of actors (at least to the ones which are still alive).
+#[derive(Debug)]
+pub struct GroupWeakAddr<A: Actor> {
+    set: PtrWeakHashSet<WeakTableWeakAddrWrap<A>>,
+}
+
+impl<A: Actor> Clone for GroupWeakAddr<A> {
+    fn clone(&self) -> GroupWeakAddr<A> {
+        GroupWeakAddr {
+            set: self.set.clone(),
+        }
+    }
+}
+
+pub struct GroupWeakAddrIterator<'a, A: Actor>(
+    weak_table::ptr_weak_hash_set::Iter<'a, WeakTableWeakAddrWrap<A>>,
+);
+
+impl<'a, A: Actor> core::iter::Iterator for GroupWeakAddrIterator<'a, A> {
+    type Item = Addr<A>;
+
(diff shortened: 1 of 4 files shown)


alvardes:
Can I get a review @Jonathas-Conceicao?

Jonathas-Conceicao:
@alvardes I'll try to find some time to carefully review this latter today.

src/address/group.rs
```diff
@@ -0,0 +1,214 @@
...
+    fn view(&self) -> Option<Self::Strong> {
+        self.upgrade().map(|addr| WeakTableAddrWrap(addr))
+    }
+}
+
+impl<A: Actor> Clone for WeakTableWeakAddrWrap<A> {
+    fn clone(&self) -> WeakTableWeakAddrWrap<A> {
+        WeakTableWeakAddrWrap(self.0.clone())
+    }
+}
```
Jonathas-Conceicao, line 47: Looks like this implementation can be derived. Less code to maintain is generally good.

src/address/group.rs
```diff
@@ -0,0 +1,214 @@
...
+pub struct GroupWeakAddr<A: Actor> {
+    set: PtrWeakHashSet<WeakTableWeakAddrWrap<A>>,
+}
+
+impl<A: Actor> Clone for GroupWeakAddr<A> {
+    fn clone(&self) -> GroupWeakAddr<A> {
+        GroupWeakAddr {
+            set: self.set.clone(),
+        }
+    }
```
Jonathas-Conceicao, line 62: Use derive from this too.

src/address/group.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+impl<A: Actor> PartialEq for GroupWeakAddr<A> {
+    fn eq(&self, other: &Self) -> bool {
+        self.set == other.set
+    }
+}
+
+impl<A: Actor> Eq for GroupWeakAddr<A> {}
```
Jonathas-Conceicao, line 92: These can be derived too.

Jonathas-Conceicao:
The general idea makes sense to me, but I'm not particularly good with names to help, comments from other contributors would be welcome.

Adding tests for this new tool is also important so we avoid any regression bugs, and they might work as a usage example as well. Also please update the CHANGES.md to mention the new struct.

alvardes:
Actually while trying to write tests for this I realized that this doesn't seem to work. :(. I tried to make this work but I don't have a lot of free time to figure out whats wrong right now so I will close this for now.

fakeshadow:
Please let me know if I get your idea. I believe something like this would be a suitable solution for your use case.
```rust
use std::any::{Any, TypeId};
use std::collections::HashMap;

use actix::prelude::*;
use actix_web::{get, App, HttpServer};

struct Host {
    children: HashMap<TypeId, Vec<Box<dyn Any>>>,
}

impl Host {
    fn add<M>(&mut self, recipient: Recipient<M>)
    where
        M: Message + Send + 'static,
        M::Result: Send,
    {
        let id = TypeId::of::<M>();

        self.children
            .entry(id)
            .or_insert(vec![])
            .push(Box::new(recipient));
    }
}

impl Actor for Host {
    type Context = Context<Self>;
}

#[derive(Default)]
struct Child1;

impl Actor for Child1 {
    type Context = Context<Self>;
}

#[derive(Default)]
struct Child2;

impl Actor for Child2 {
    type Context = Context<Self>;
}

#[derive(Clone)]
struct Broadcast;

#[derive(Clone)]
struct Register<M>(Recipient<M>)
where
    M: Message + Send + 'static,
    M::Result: Send;

impl Message for Broadcast {
    type Result = ();
}

impl<M> Message for Register<M>
where
    M: Message + Send + 'static,
    M::Result: Send,
{
    type Result = ();
}

impl<M> Handler<Register<M>> for Host
    where
        M: Message + Send + 'static,
        M::Result: Send,
{
    type Result = ();

    fn handle(&mut self, msg: Register<M>, _ctx: &mut Context<Self>) -> Self::Result {
        self.add(msg.0);
    }
}

impl Handler<Broadcast> for Host {
    type Result = ();

    fn handle(&mut self, msg: Broadcast, _ctx: &mut Context<Self>) -> Self::Result {
        self.children
            .get_mut(&TypeId::of::<Broadcast>())
            .unwrap()
            .into_iter()
            .for_each(|obj| {
                obj.downcast_mut::<Recipient<Broadcast>>()
                    .unwrap()
                    .do_send(msg.clone())
                    .unwrap()
            });
    }
}

impl Handler<Broadcast> for Child1 {
    type Result = ();

    fn handle(&mut self, _msg: Broadcast, _ctx: &mut Context<Self>) -> Self::Result {
        println!("Child1: we got message from master")
    }
}

impl Handler<Broadcast> for Child2 {
    type Result = ();

    fn handle(&mut self, _msg: Broadcast, _ctx: &mut Context<Self>) -> Self::Result {
        println!("Child2: we got message from master")
    }
}

#[actix_rt::main]
async fn main() -> std::io::Result<()> {
    let master = Host::create(|_| {
        Host {
            children: Default::default()
        }
    });

    let child1 = Child1::start_default();

    let child2 = Child2::start_default();

    master.send(Register(child1.recipient())).await.unwrap();

    master.send(Register(child2.recipient())).await.unwrap();

    master.send(Broadcast).await.unwrap();

    actix_rt::time::delay_for(std::time::Duration::from_secs(1)).await;

    Ok(())
}
```
By using Recipent instead of Addr you can have multiple children actors implement different kinds of messages. 
Yes there is some runtime cost for an extra layer of trait object and there is no WeakRecipient but I believe they are not hard to implement.

PR #407, opened 2020-07-09
Make `Request` `Send` regardless of actor type

### LINKED ISSUE #405: Implement `Send` for `actix::message::Request` explicitly
## Expected Behavior
`actix::message::Request` explicitly (and unconditionally) implements `Send` regardless if it's `A` type-argument implements `Send`.

## Current Behavior

- `Request` contains `PhantomData<A>`
- `PhantomData<T>: Send` only when `T: Send`
- so, as a result, `Request` only implements `Send`, when it's `A` type-argument implements `Send`

As far as I can see, there is no real reason for that. `Request` does not ever directly interact with actor, only sends a message through a channel.

## Possible Solution

I've looked through the code, this seems to be enough:

```rust
unsafe impl<A, M> Send for Request<A, M> 
where
    A: Handler<M>,
    A::Context: ToEnvelope<A, M>,
    M: Message + Send,
    M::Result: Send,
{
}
```

## Context

I'm working on a web-service using `actix-web` and `juniper` (GraphQL implementation in Rust). Due to current shortcomings in `juniper` it requires `Send` futures. So, if I want to interact with my actors from inside my GraphQL handlers, I have to heavily bound all my actors to be `Send`.

## Your Environment

- Rust Version: 1.44.1
- Actix Version: 0.9, 0.10-*

ffuugoo:
Removes `Request`'s `PhantomData<A>` field, as `A` is already constrained in `Request` (there is a `AddressSender<A>` field) and `PhantomData` just seems to be overlooked during previous `Request` refactoring.

<details>
<summary>HISTORY</summary>

Changes `Request`'s `PhantomData<A>` to `PhantomData<AtomicPtr<A>>` (which is always `Send` and `Sync`), which makes `Request` auto implement `Send` and `Sync` regardless if `A` is `Send` or `Sync`.

</details>

Fixes #405.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@
 [#384]: https://github.com/actix/actix/pull/384
 [#403]: https://github.com/actix/actix/pull/403
 [#404]: https://github.com/actix/actix/pull/404
+[#407]: https://github.com/actix/actix/pull/407
 
 ## [0.10.0-alpha.3] - 2020-05-13
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@
 
 * Update `parking_lot` to 0.11 [#404]
 
+* Remove unnecessary `PhantomData` field from `Request` making it `Send + Sync`
+  regardless if `Request`'s type-argument is `Send` or `Sync` [#407]
+
 [#384]: https://github.com/actix/actix/pull/384
 [#403]: https://github.com/actix/actix/pull/403
 [#404]: https://github.com/actix/actix/pull/404
--- a/src/address/message.rs
+++ b/src/address/message.rs
@@ -1,5 +1,4 @@
 use std::future::Future;
-use std::marker::PhantomData;
 use std::pin::Pin;
 use std::task::{self, Poll};
 use std::time::Duration;
@@ -27,7 +26,6 @@ where
     rx: Option<oneshot::Receiver<M::Result>>,
     info: Option<(AddressSender<A>, M)>,
     timeout: Option<Delay>,
-    act: PhantomData<A>,
 }
 
 impl<A, M> Request<A, M>
@@ -44,7 +42,6 @@ where
             rx,
             info,
             timeout: None,
-            act: PhantomData,
         }
     }
 


cetra3:
This smells weird to me!  Shouldn't this work already if `A` is `Send`? Why would `A` not be `Send` & If so is it safe to do this? I.e, we might be changing an invariant used in `unsafe` code somewhere else.

ffuugoo:
> This smells weird to me! Shouldn't this work already if `A` is `Send`? Why would `A` not be `Send` & If so is it safe to do this? I.e, we might be changing an invariant used in `unsafe` code somewhere else.

Well, it is `Send` if `A` is `Send`, but that's sort of the whole point. Actix ___allow___ you to create and spawn non-`Send` actors. And it's not that hard to imagine, that if actor is spawned on the same thread and then never rescheduled to another one, then sure, it does not have to be `Send`. `Actor` trait signatures kinda suggest that is the case: `Actor::start` does not require `Send` and `Actor::start_in_arbiter` accepts a closure that produces your actor and only the closure itself required to be `Send` (i.e., exactly to avoid actor being `Send` and to produce it directly on the thread it will be run at).

Of course I can miss something, I am no actix expert, but so far I fail to see how `Request` can depend on `A` being `Send`/non-`Send`: it's just a few channels and a message bundled together. It does not interact with actor in any way except for sending a message into the channel and receiving result from another one.

ffuugoo:
Now that I've written all of this, I've realized that `Request` does not even need to have `PhantomData`? It has an `AddressSender<A>` field, which means `A` is already constrained and `PhantomData` is obsolete.

robjtede:
Great, can you add a changelog entry, @ffuugoo ?

ffuugoo:
> Great, can you add a changelog entry, @ffuugoo ?

Sure.

ffuugoo:
@robjtede Done.

CHANGES.md
```diff
@@ -10,6 +10,9 @@
 
 * Update `parking_lot` to 0.11 [#404]
 
+* Remove unnecessary `PhantomData` field from `Request` making it `Send + Sync`
+  regardless if `Request`'s type-argument is `Send` or `Sync` [#407]
```
robjtede, line 14: the PR link needs an entry below, it doesnt auto-link unfortunately

ffuugoo: Oh wow, some proper tunnel vision on my part here. Sorry. Fixed.

PR #411, opened 2020-07-20
Check code with rustfmt not to introduce format commits

JohnTitor:
## PR Type

Chore

## Overview

Currently, we don't force contributors to format their PRs. I think it's worth checking code with rustfmt on CI since an unspecified number of people may touch the code and formatting commit is noisy to follow commit history.
I use stable rustfmt here for two reasons:
- nightly rustfmt enables unstable rules.
- nightly rustfmt is sometimes broken as it's managed as submodule rather than subtree like Clippy.

### DIFF
--- a/.github/workflows/clippy-fmt.yml
+++ b/.github/workflows/clippy-fmt.yml
@@ -2,18 +2,31 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
-name: Clippy Check
+name: Clippy and rustfmt Check
 jobs:
   clippy_check:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+
+      - uses: actions-rs/toolchain@v1
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/clippy-fmt.yml
```diff
@@ -13,7 +13,17 @@ jobs:
...
           args: --all-features --all --tests
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          components: rustfmt
+          override: true
+      - name: Check with rustfmt
+        uses: actions-rs/cargo@v1
+        with:
+          command: fmt --all -- --check
```
robjtede, line 29: Since this is a much faster check than clippy, I'd say put it first it fails quicker and doesn't install crates.

JohnTitor: Indeed! Will do so.

PR #418, opened 2020-09-06
add `WeakRecipient<M>`

hoodie:
## PR Type

Feature

## PR Checklist

## Overview

Hi there,
I'm trying to convert my project to make greater use of `Recipient<M>` instead of `Addr<A>` since I believe it allows to loosen some tight coupling here and there. Unfortunately there is no equivalent to `WeakAddr<A>` for Recipients yet, so I cannot really use Recipents at the moment without running the risk of leaking objects.
Therefore I would like to suggest adding `WeakRecipients` to actix.

Thank you

### DIFF
--- a/src/address/channel.rs
+++ b/src/address/channel.rs
@@ -32,6 +32,8 @@ where
 
     fn boxed(&self) -> Box<dyn Sender<M> + Sync>;
 
+    fn arced(&self) -> Arc<dyn Sender<M> + Sync>;
+
     fn hash(&self) -> usize;
 
     fn connected(&self) -> bool;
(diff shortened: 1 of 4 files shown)


src/address/channel.rs
```diff
@@ -468,6 +470,9 @@ where
     fn boxed(&self) -> Box<dyn Sender<M> + Sync> {
         Box::new(self.clone())
     }
+    fn arced(&self) -> Arc<dyn Sender<M> + Sync> {
```
jplatte, line 473:
```suggestion
    }

    fn arced(&self) -> Arc<dyn Sender<M> + Sync> {
```

hoodie:
Hi again!
I could have closed this a while ago, sorry about that, but this implementation does not work.
In this one I create an `Arc` inside the the recipient and therefor my `WeakRecipient` is invalid as soon as the `Recipient` goes away. I'm about to open a new version of this PR.

PR #421, opened 2020-09-11
Remove the Unpin requirement on Response::fut()

Michael-F-Bryan:
# PR Type

This is a *Feature* -  it allows users to write code that they previously couldn't due to an overly strict `where` clause.

## PR Checklist

- [X] Tests for the changes have been added / updated.
- [X] Documentation comments have been added / updated.
- [X] Format code with the latest stable rustfmt

## Overview

This PR removes a the `Unpin` requirement from the `Response::fut()` constructor's `where` clause, allowing a `Response` to be constructed using an `async` block or result of an `async` function. 

Previously, a `Response` could only be constructed via combinators (`futures::FutureExt` and friends) and using an `async` block would result in a type error.

This should be fully backwards compatible because: 

1. Any type that is `T: Future + Unpin + 'static` is also `T: Future + 'static`
2. The user can't directly access the `ResponseTypeItem::Fut` variant. So they won't know it's changed from `Box<dyn Future>` to `Pin<Box<dyn Future>>`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,10 +13,13 @@
 * Remove unnecessary `PhantomData` field from `Request` making it `Send + Sync`
   regardless if `Request`'s type-argument is `Send` or `Sync` [#407]
 
+* The `where` clause on `Response::fut()` was relaxed to no longer require `T: Unpin`, allowing a `Response` to be created with an `async` block [#421]
+
 [#384]: https://github.com/actix/actix/pull/384
 [#403]: https://github.com/actix/actix/pull/403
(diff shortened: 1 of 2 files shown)


Michael-F-Bryan:
Is this large enough to warrant a change log entry?

JohnTitor:
> Is this large enough to warrant a change log entry?

I'd love to see an entry for this :)

JohnTitor (approved):
I think this is reasonable.

Michael-F-Bryan:
> I'd love to see an entry for this :)

Done.

JohnTitor:
Thanks!

PR #425, opened 2020-09-15
Provide a better error message when mocking fails and improve ergonomics

Michael-F-Bryan:
## PR Type

Refactor. Optionally with some additional functionality.

## PR Checklist

## Overview

The `Mocker` type passes around `Box<dyn Any>` for both messages and message results. That means it's not uncommon for your `mock` function to accidentally return the wrong type and panic with a rather cryptic `"wrong return type for message"`. 

The main feature of this PR is to adjust the panic message to mention the type `M: Message` the `Mocker<T>` is handling, and the type `M::Result` we expected the `mock` function to return.

As an addition (which can be cherry picked out if we don't want to include it), the implementation of `Handler<M>` required that  instead of returning a `M::Result` directly, the `mock` function returns a `Option<M::Result>` so we can do `downcast_mut()` and a `mem::replace()` to extract the `M::Result`. 69f4135 switches to using `ret.downcast::<M::Result>()` instead of `downcast_mut()` so we can take the `M::Result` by value, chaining on a `ret.downcast::<Option<M::Result>>()` so we retain the original behaviour.

A second optional addition, 5d47d1f, adds a `Mocker::mock_one()` constructor for when your `Mocker` will only need to mock a single message. This lets the user avoid writing code for manually boxing and downcasting.

This change introduces the use of `std::any::type_name()` which only became stable in Rust 1.38.0. The MSRV for Actix is 1.40, so this PR shouldn't introduce any backwards compatibility issues.

### DIFF
--- a/src/actors/mocker.rs
+++ b/src/actors/mocker.rs
@@ -73,17 +73,24 @@ where
 {
     type Result = M::Result;
     fn handle(&mut self, msg: M, ctx: &mut Self::Context) -> M::Result {
-        let mut ret = (self.mock)(Box::new(msg), ctx);
+        let ret = (self.mock)(Box::new(msg), ctx);
 
-        match ret.downcast_mut::<Option<M::Result>>() {
-            Some(response) => response.take().unwrap(),
-            None => {
-                panic!(
-                    "Wrong return type for message. Expected a {} when responding to {}",
-                    any::type_name::<M::Result>(),
-                    any::type_name::<M>(),
-                );
-            }
+        match ret.downcast::<M::Result>() {
+            // The result was returned directly.
+            Ok(response) => *response,
+            // Note: We also handle the case where Option<M::Result> is returned
+            // for backwards compatibility.
+            Err(other) => match other.downcast::<Option<M::Result>>() {
+                Ok(mut response) => response.take().expect("No response provided"),
+                // We got something else, blow up with a useful error message.
+                Err(_) => {
+                    panic!(
+                        "Wrong return type for message. Expected a {} when responding to {}",
+                        any::type_name::<M::Result>(),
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
As a note, I couldn't quite get the `#[should_panic(expected = "...")]` part of the test to work. It looks like the panic occurs on a tokio thread and the original thread only gets a "mailbox closed" error.

```
---- actors::mocker::tests::users_get_a_useful_panic_message_when_mocking_fails stdout ----
thread 'actors::mocker::tests::users_get_a_useful_panic_message_when_mocking_fails' panicked at 'Wrong return type for message. Expected a actix::actors::mocker::tests::Pong when responding to actix::actors::mocker::tests::Ping', src\actors\mocker.rs:101:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'actors::mocker::tests::users_get_a_useful_panic_message_when_mocking_fails' panicked at 'called `Result::unwrap()` on an `Err` value: MailboxError(Mailbox has closed)', src\actors\mocker.rs:158:60
Panic in Arbiter thread.
note: panic did not contain expected string
      panic message: `"called `Result::unwrap()` on an `Err` value: MailboxError(Mailbox has closed)"`,
 expected substring: `"Wrong return type for message. Expected a actix::actors::mocker::tests::Pong when responding to actix::actors::mocker::tests::Ping"`

failures:
    actors::mocker::tests::users_get_a_useful_panic_message_when_mocking_fails

test result: FAILED. 4 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

robjtede:
Marking as draft due to CI failures. LMK if you're still interested in pursuing this PR.

PR #432, opened 2020-10-09
Add WeakRecipient (v2)

hoodie:
With this change you can now call `.recipient()` on `WeakAddr<A>` just
like on `Addr<A>`. In this case you get a `WeakRecipient<M>` that can
be upgraded to a `Recipient<M>`.

## PR Type

INSERT_PR_TYPE

## PR Checklist

## Overview
Hi there,
I'm trying to convert my project to make greater use of `Recipient<M>` instead of `Addr<A>` since I believe it allows to loosen some tight coupling here and there. Unfortunately there is no equivalent to `WeakAddr<A>` for Recipients yet, so I cannot really use Recipients at the moment without running the risk of leaking objects.
Therefore I would like to suggest adding `WeakRecipients` to actix.

This PR supersedes #418, which was flawed. In that attempt I made the `Box` inside the `Recipient` an `Arc` and downgraded it.
you would call
```rust
let weak_rcp = my_actor.addr().recipient().downgrade();
```

That however meant  that you the `WeakRecipient` was invalidated immediately.
In this second attempt I'm creating the `WeakRecipient` from an existing `WeakAddr`. 
```rust
let weak_rcp = my_actor.addr().downgrade().recipient();
```

Thank you very much for your review.

### DIFF
--- a/src/address/channel.rs
+++ b/src/address/channel.rs
@@ -584,7 +584,6 @@ where
     fn boxed(&self) -> Box<dyn WeakSender<M> + Sync> {
         Box::new(self.clone())
     }
-
 }
 
 //
--- a/src/address/channel.rs
+++ b/src/address/channel.rs
@@ -45,6 +45,7 @@ where
     ///
     /// Returns [`None`] if the actor has since been dropped.
     fn upgrade(&self) -> Option<Box<dyn Sender<M> + Sync>>;
+    fn boxed(&self) -> Box<dyn WeakSender<M> + Sync>;
 }
 
 /// The transmission end of a channel which is used to send values.
@@ -579,6 +580,11 @@ where
             None
         }
     }
+
+    fn boxed(&self) -> Box<dyn WeakSender<M> + Sync> {
+        Box::new(self.clone())
+    }
+
 }
 
 //
--- a/src/address/channel.rs
+++ b/src/address/channel.rs
@@ -36,6 +36,13 @@ where
 
     fn connected(&self) -> bool;
 }
+pub(crate) trait WeakSender<M>: Send
+where
+    M::Result: Send,
+    M: Message + Send,
+{
+    fn upgrade(&self) -> Option<Box<dyn Sender<M> + Sync>>;
+}
 
 /// The transmission end of a channel which is used to send values.
 ///
@@ -551,6 +558,22 @@ impl<A: Actor> WeakAddressSender<A> {
         match Weak::upgrade(&self.inner) {
             Some(inner) => Some(AddressSenderProducer { inner }.sender()),
             None => None,
+        }
+    }
+}
+
+impl<A, M> WeakSender<M> for WeakAddressSender<A>
+where
+    A: Handler<M>,
+    A::Context: ToEnvelope<A, M>,
+    M::Result: Send,
+    M: Message + Send + 'static,
+{
+    fn upgrade(&self) -> Option<Box<dyn Sender<M> + Sync>> {
+        if let Some(inner) = WeakAddressSender::upgrade(&self) {
+            Some(Box::new(inner))
+        } else {
+            None
         }
     }
 }
--- a/src/address/channel.rs
+++ b/src/address/channel.rs
@@ -41,6 +41,9 @@ where
     M::Result: Send,
     M: Message + Send,
 {
+    /// Attempts to upgrade a `WeakAddressSender<A>` to a [`Sender<M>`]
+    ///
+    /// Returns [`None`] if the actor has since been dropped.
     fn upgrade(&self) -> Option<Box<dyn Sender<M> + Sync>>;
 }
 
--- a/src/address/mod.rs
+++ b/src/address/mod.rs
@@ -348,6 +348,16 @@ where
     wtx: Box<dyn WeakSender<M> + Sync>,
 }
 
+impl<M> fmt::Debug for WeakRecipient<M>
+where
+    M: Message + Send,
+    M::Result: Send,
+{
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(fmt, "WeakRecipient {{ /* omitted */ }}")
+    }
+}
+
 impl<M> WeakRecipient<M>
 where
     M: Message + Send,
@@ -358,7 +368,7 @@ where
     }
 
     /// Attempts to upgrade the `WeakRecipient<M>` pointer to an `Recipient<M>`, similar to `WeakAddr<A>`
-    pub fn upgrade(self) -> Option<Recipient<M>> {
+    pub fn upgrade(&self) -> Option<Recipient<M>> {
         self.wtx.upgrade().map(Recipient::new)
     }
 }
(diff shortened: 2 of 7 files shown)


hoodie:
@JohnTitor @robjtede ?

src/address/mod.rs
```diff
@@ -329,6 +339,58 @@ where
...
+    A: Handler<M>,
+    M::Result: Send,
+    A::Context: ToEnvelope<A, M>,
+{
+    fn into(self) -> WeakRecipient<M> {
+        WeakRecipient::new(Box::new(self.wtx))
+    }
+}
+
+/// The `Recipient` type allows to send one specific message to an
```
lucab, line 388: Spurious docstring (possibly a copy-paste leftover)?

src/address/channel.rs
```diff
@@ -36,6 +36,13 @@ where
 
     fn connected(&self) -> bool;
 }
+pub trait WeakSender<M>: Send
```
lucab, line 39: Would `pub(crate)` be enough here?

src/address/mod.rs
```diff
@@ -329,6 +339,58 @@ where
     }
 }
 
+/// A weakly referenced counterpart to `Recipient<M>`
```
lucab, line 342: As a user, I'd find some more docstring helpful here to explain what's the typical usecase of this weak recipient.
In particular, I found your mention of "risk of leaking objects" enlightening; it could mirrored and expanded in the docstring here.

lucab (commented):
(just some casual PR reading, I'm not a maintainer here)

src/address/mod.rs
hoodie: whoops, yes indeed, I was shuffling files around a bit there, after I had moved recipient and weak_recipient into their own submodules and then reintegrated them. these files contain a bunch of structs at once and tons of trait implementations, I would at some point ask if we perhaps want to split them up a little.

tanks for the find!

src/address/mod.rs
hoodie: this doc-string is a reflection of the one for `WeakAddr`. You are right about the usecase example, IDK if it fits in here, but maybe an example or integration test would be good.

hoodie:
Ping @JohnTitor 😀

JohnTitor (commented):
Hey, sorry for the delay, and thanks for pinging!
The approach should be reasonable and the change looks good to me. But I'm going to cc @Jonathas-Conceicao for further reviewing, if you're available now :)
Also, could you add an entry to the changelog?

actix-derive/src/message.rs
```diff
@@ -69,10 +69,10 @@ fn get_attribute_type_multiple(
             .map(|m| meta_item_to_ty(m).ok())
             .collect())
     } else {
-        return Err(syn::Error::new_spanned(
+        Err(syn::Error::new_spanned(
             attr,
             format!("The correct syntax is #[{}(type, type, ...)]", name),
-        ));
+        ))
```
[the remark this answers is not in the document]
hoodie: sure, will do 👍

src/address/mod.rs
JohnTitor: The docs enhancement isn't absolutely necessary, but it'd be great if we could add a minimal example for it :)

hoodie:
@robjtede why did you merge `master` into this branch?

robjtede:
@hoodie repo has a linear history policy, so if this is accepted it needs to be based on HEAD for CI to check it properly before squashing

hoodie:
ah ok, so merging master in here is a actually better than rebasing onto master for this use case, get it

robjtede:
@hoodie either is appropriate and just as good, many people don't like rebasing and it's just a button on github to merge master in

src/address/channel.rs
```diff
@@ -555,6 +562,22 @@ impl<A: Actor> WeakAddressSender<A> {
...
 }
 
+impl<A, M> WeakSender<M> for WeakAddressSender<A>
+where
+    A: Handler<M>,
+    A::Context: ToEnvelope<A, M>,
+    M::Result: Send,
+    M: Message + Send + 'static,
+{
+    fn upgrade(&self) -> Option<Box<dyn Sender<M> + Sync>> {
```
Jonathas-Conceicao, line 572: I believe we should have a doc note about when this `Option` would be `None`, either here or at the trait's definition.

tests/test_address.rs
```diff
@@ -100,7 +101,51 @@ impl Actor for WeakRunner {
...
+
+        let weak1 = addr1.downgrade().recipient();
+        let weak2 = addr2.downgrade().recipient();
+        // not strictly necessary, weak1 is dropped by RIAA since nobody in `run_later` uses it.
+        // to make this test red, move drop(addr1) to the line before `System::current().stop()`
+        drop(addr1);
+
+        ctx.run_later(Duration::new(0, 1000), move |_, _| {
+            if weak1.upgrade().is_none() {
+            } else {
```
Jonathas-Conceicao, line 130:
```suggestion
            if weak1.upgrade().is_some() {
```

Jonathas-Conceicao (changes requested):
The feature proposal seams good overall, I left a couple of comments for some minor improvements.

src/address/channel.rs
hoodie: sure, but probably at the definition of the trait and not here right?

Jonathas-Conceicao: Both will be generated to the doc, but I think the trait it might be more noticeable. So if the `None` has a generic cause it would be better to have it there, if it's specific to the type's implementation we can have it here.

hoodie: I added it on the trait, but it's `pub(crate)`, so it won't show up in any docs anyway.
Plus: the semantics are basically the same as for `Weak<T>::upgrade()`.
Alright by you @Jonathas-Conceicao ?

hoodie: I added it on the trait, but it's pub(crate), so it won't show up in any docs anyway.
Plus: the semantics are basically the same as for Weak<T>::upgrade().
Alright by you @Jonathas-Conceicao ?

hoodie:
@JohnTitor I have added an example for `WeakAddr` and `WeakRecipient`.
In the process I noticed that the signature of `Recipient::do_send()` is not consistent with the signature of `Addr::do_send()`, I believe the former should not return a `Result<...>`. I just opened #440 to track that.

hoodie:
Ok, since this has gotten a bunch of changes since I opened it originally I thought I'd give a quick recap since I know it's a bit hard to follow these things:

* [x] added `WeakRecipient`
* [x] added trait doc for `WeakAddressSender`
* [x] added two new examples
* [ ] make clippy happy

any other requests?

src/address/mod.rs
lucab: @hoodie thanks for the pair of examples, much appreciated!

hoodie:
ping @Jonathas-Conceicao

Jonathas-Conceicao (approved):
lgtm, thanks for working on this!

JohnTitor:
Thanks all!

PR #441, opened 2020-10-20
Remove return value from Recipient::do_send

### LINKED ISSUE #440: Signatures of `Recipient::do_send()` and `Addr::do_send()` are not consistent
## Expected Behavior
`Recipient::do_send()` should not return a `Result`, same as `Addr::do_send()` does.

## Current Behavior
Currently `Recipient::do_send()` is identical in behavior with `Recipient::try_send()`, this seems to be an oversight.

## Possible Solution
Consume the Result inside the method body, same as `Addr::do_send()` does:
https://github.com/actix/actix/blob/master/src/address/mod.rs#L99

## Context
This is a breaking API change, but a consistency improvement, users of `Recipient::do_send()` presumably already throw away the Result.

hoodie:
do_send() is supposed to be the fire and forget way of sending, as
opposed to try_send(). This change aligns the signatures of
`Addr::do_send` and `Recipient::do_send()`.

## PR Type

Refactor

## PR Checklist

## Overview

Closes #440

### DIFF
--- a/src/address/mod.rs
+++ b/src/address/mod.rs
@@ -236,8 +236,9 @@ where
     /// Sends a message.
     ///
     /// Deliver the message even if the recipient's mailbox is full.
-    pub fn do_send(&self, msg: M) -> Result<(), SendError<M>> {
-        self.tx.do_send(msg)
+    /// If the mailbox is closed, the message is silently dropped.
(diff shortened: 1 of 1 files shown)


robjtede:
included this in #525 (0.13)

Issue #438, opened 2020-10-17
async handlers with responses


mankinskin:
Are there any plans for adding something like an `AsyncHandler` trait:
```rust
pub trait AsyncHandler<M> 
where
    Self: Actor,
    M: Message, 
{
    type Result: MessageResponse<Self, M>;
    async fn handle(&mut self, msg: M, ctx: &mut Self::Context) -> Self::Result;
}
```
which can return a future and has internally managed async memory access?

I would really like to use actix with async APIs. I can use it with [actix-interop](https://github.com/Diggsey/actix-interop) right now, but that introduces a lot of boilerplate, especially because it requires the state used by the async code to be completely `'static`, which means I have to manage all of my actor's state (that should be accessed by async functions) in a `lazy_static` variable. It would be nice if this could be handled by actix and if I could just write async code directly in an actix handler, which can also return responses.

fakeshadow:
Rust doesn't support async methods in trait so what you write is wrong in syntex.

That said It's likey that you are talking about using macros like `asynct_trait` but there are downsides:

1. It's desugar to `Pin<Box<dyn Future>>` which means you get what we have now in the end. It's an illusion that you are writing async trait method. It sure is useful but is it worth it to complicate the code with such illusion?

2. The macro would have to figure a way to put part of your handle method in async block as you can't borrow self in an async block. It's the design of actix and how an actix actor can poll multiple messages concurrently. This result in either like `actix-interop` using static and locks to aviod the lifetime issue and mutable borrow issue. Or the macro would do a very complex logic to figure out which part of your handle method can be put into an async block before throwing an error if it can not.  

What you actually want is an actor patter rather than actix. You want an actor that handle message one by one in async manner. It can borrow itself in async method and all other messages are held for later until the future is finished.

mankinskin:
> 1. It's desugar to `Pin<Box<dyn Future>>` which means you get what we have now in the end. It's an illusion that you are writing async trait method. It sure is useful but is it worth it to complicate the code with such illusion?

If by illusion you mean syntax sugar, then that would make it less complicated wouldn't it? It would be much more convenient not to have to worry about boxing and pinning the future manually.

> What you actually want is an actor patter rather than actix

I thought actix implements an actor pattern.. I was thinking actix can store the actor's state in a static variable and execute the async handler functions in a tokio executor and handle all of the messaging and memory access internally.

[...]

Issue #378, opened 2020-04-14
Async/await interop suggestion


Diggsey:
I wrote a [small crate](https://github.com/Diggsey/actix-interop) for using async/await syntax with actix. It also makes the "AtomicResponse" type obsolete by providing a more ergonomic mechanism to achieve the same thing. I think it would be nice to have as part of actix itself.

The basic idea is to use scoped TLS (at least until [custom future contexts](https://github.com/rust-lang/rfcs/issues/2900) are supported) to store the actor and its context.

There are three functions:

- `FutureInterop::interop_actor()`

    This is equivalent to the `into_actor()` method except that it populates the TLS slots with the actor and the context before polling the inner future. There's also a boxed variant of this function.

- `with_ctx`

    This takes a closure and immediately calls it, passing in mutable references to the actor and context obtained from the TLS slots to that closure.

- `critical_section`

    This takes a future/async block and uses `ctx.wait()` to make sure it has exclusive access to the actor state while it is running. It returns a future itself so you can seamlessly demarcate blocks of code within an async/await block as being critical sections.

Example:

```rust

impl Handler<Request> for MyActor {
    type Result = ResponseActFuture<Self, Result<Response, ()>>;

    fn handle(&mut self, msg: Request, _ctx: &mut Context<Self>) -> Self::Result {
        async move {
            // Modify the actor state in some way
            with_ctx(|a: &mut Self, _| a.mutate());

            ... <some async code> ...

            // Run some code with exclusive access to the actor
            let result = critical_section::<Self, _>(async {
                // Put the actor into an inconsistent state.
                // We don't want this visible outside the critical section.
                with_ctx(|a: &mut Self, _| a.begin_mutation());

                ... <some async code> ...

                // Make everything consistent again
                Ok(with_ctx(|a: &mut Self, _| a.end_mutation()))
            }).await;

            ... <some async code> ...

            // Return a result
            result
        }.interop_actor_boxed(self)
    }
}
```

Some things which are very hard to do right now are made much easier in this pattern (for example, when only half of a message handler needs to run atomically). Also, all of the normal futures combinators can be used, rather than the more limited set exposed under `actix::fut`.

Currently the crate relies on a patch to `scoped_thread_local` which hasn't been merged yet, which is why I haven't published it.

Diggsey:
This crate is now published to crates.io.

david-mcgillicuddy-moixa:
[...]

PR #449, opened 2020-12-01
add atomic unsafe response type. enable async handler

### LINKED ISSUE #308: Use of actor reference on Async Actor handle

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

AtomicResponse actually could enable borrowing Actor and it's Context in async/await.
This PR just use it with some unsafe code to enable async/await.

It's also possible to immutably borrowing Actor and it's Context in async/await without atomic,
by grouping shared reference responses and block actor on them.
It's basically a batched version of AtomicResponse. 

Closes #308 #378 #438


fakeshadow:
I'm not yet decided how to guard the input. 
Introducing async_trait for a future proof of Handler trait is what I prefer but it could mean harm to SyncActor usage.
Use an extra closure would not break but with more boilerplate

robjtede:
So what's the path forward for this PR. Is it ready for review and if so which parts should be focussed on.

A note after skimming over a couple files: a few fns are documented with "do not expose to public api". Are these fns better as unsafe? Also consider documenting the invariants that must be upheld regardless.

fakeshadow:
Hard to say it's ready as it's pretty hard to test the unsafes added. It would need some really edge cases to see if it holds up.

`Context::wait_concurrent` is just a method to push 'static lifetime future to actor's context. It's like `actix_rt::spawn` on an actor and they are safe. It' just not suggested to be used in public API. 

The unsafe code only affect people who actually call `ResponseAsync` in handler. For people avoid using it there is no actual change to the code base other than a few more `Vec::len` check against an empty vector.

fakeshadow:
Making this a draft for now until proper tests are added.

fakeshadow:
I have reduced most changes related to actor context. Changing this PR to an `AtomicResponse` similar feature that uses unsafe to permit borrowing actor and context in async block. 
It's supposed to work exactly like `AtomicResponse` but async/await friendly.

robjtede:
definitely excited about this but i think we should do a release without it first

fakeshadow:
Agreed. I'm not confident in this either.

There are still small issues to this approach so it's more like an experiment for the Concurrent variant AsyncResponse.

fakeshadow:
Current issue I find on this approach:

`Actor::stopping` have access to actor and context with mut reference. So an actor with living waiting async response would find the state could be modified between yield points where the actor goes through the stopping and back to running state with modifed state. This is not a double mut borrow by any means but would still be annoying as the actor state loses consistency between await point when above situation happen.

The only way to fix this is to change the current behavior of handling stopping or use a spearate queue for async responses where they are thrown away when actor goes into stopping state.

robjtede:
lmk if you want a review of this now

fakeshadow:
@robjtede I think this is ready for review.

fakeshadow:
I don't have time to keep this PR up to date anymore. It's welcomed to step in and take charge of it.

actix/src/handler.rs
```diff
@@ -151,6 +151,93 @@ where
...
+where
+    A: Actor,
+    A::Context: AsyncContext<A>,
+    M: Message<Result = T>,
+    T: 'static,
+{
+    fn handle(self, ctx: &mut A::Context, tx: Option<OneshotSender<M::Result>>) {
+        match self {
+            AsyncResponse::Atomic(fut) => {
+                ctx.wait(fut.map(move |res, _, _| tx.send(res)));
```
thalesfragoso, line 232: Sorry for reviving this....

Isn't this unsound already ? You have two `&mut A::Context` alive at the same time, one given to the method and the other inside the future.

PR #453, opened 2020-12-31
use pin_project_lite. remove futures-util

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Replace `futures_channel` with `tokio::sync`
Replace `pin_project` with `pin_project_lite`
Replace `futures_util` with `futures_core/sink/task`

Sadly this does not cut down the amount of deps with default featurs.(5 less dep with no-default-features)


CHANGES.md
```diff
@@ -28,6 +28,9 @@
   `Arbiter::spawn` and `actix_rt::spawn` would panic outside the context of `actix::System`. They must be call
   inside `System::run`,`SystemRunner::run` or `SystemRunner::block_on`.
   More information can be found [here](https://github.com/actix/actix-net/issues/206#issuecomment-717769654) [#447]
+* `actix::fut::Either` is change to enum with field name. `Either::left` and `Either::right` function can be used
```
therealprof, line 31:
```suggestion
* `actix::fut::Either` is changed to enum with field name. `Either::left` and `Either::right` function can be used 
```

Cargo.toml
```diff
@@ -57,7 +58,7 @@ trust-dns-resolver = { version = "0.20.0" , optional = true, default-features =
 
 [dev-dependencies]
 doc-comment = "0.3"
-futures-channel = { version = "0.3.7", default-features = false, features = ["sink"] }
+futures-util = { version = "0.3.7", default-features = false }
```
therealprof, line 61:
```suggestion
futures-util = { version = "0.3.8", default-features = false }
```

I'm aware it doesn't really matter but since 0.3.8 is the current version, we might as well use that. ;)

fakeshadow: Thanks for point out but we are using 0.3.7 across all the actix-* crates.
There is no lock on the crates so users can choose to use anything above 0.3.7

therealprof (approved):
I'm not a huge fan of switching between using `pin_project_lite::pin_project!` and `using pin_project_lite::pin_project; pin_project!` but other than that looks great to me!

Cargo.toml
robjtede: Yep, no need to needlessly bound the version in a library when we don't need to. It's ^0.3.7 to avoid an security warning. Eg: https://deps.rs/crate/actix-web/3.3.2

src/fut/either.rs
```diff
@@ -22,8 +41,8 @@ impl<A, B, T> Either<(T, A), (T, B)> {
     /// Here, the homogeneous type is the first element of the pairs.
     pub fn factor_first(self) -> (T, Either<A, B>) {
```
robjtede, line 42:
```suggestion
    pub fn factor_left(self) -> (T, Either<A, B>) {
```

src/fut/either.rs
```diff
@@ -34,8 +53,8 @@ impl<A, B, T> Either<(A, T), (B, T)> {
     /// Here, the homogeneous type is the second element of the pairs.
     pub fn factor_second(self) -> (Either<A, B>, T) {
```
robjtede, line 54:
```suggestion
    pub fn factor_right(self) -> (Either<A, B>, T) {
```

src/fut/stream_finish.rs
```diff
@@ -1,21 +1,24 @@
...
+    #[derive(Debug)]
+    pub struct StreamFinish<S: ActorStream> {
+        #[pin]
+        stream: S
+    }
+}
 
-pub fn new<S: ActorStream>(s: S) -> StreamFinish<S> {
-    StreamFinish(s)
+pub fn new<S: ActorStream>(stream: S) -> StreamFinish<S> {
```
robjtede, line 20: I personally feel this style is clearer in docs for simple bounds.

```suggestion
pub fn new(stream: impl ActorStream) -> StreamFinish<S> {
```

robjtede (approved):

src/fut/stream_finish.rs
robjtede: Oh derp... sorry, S is used actually.

Cargo.toml
therealprof: I'm not saying we should fix the version, rather the opposite. I actually would even prefer going 0.3 instead of specifying the minor. But by specifying 0.3.8 we're forcing all dependencies to go at least to 0.3.8, without `cargo update` which is a good thing in my book.

robjtede: Allowing versions below 0.3.7 is a security risk and shows up on our deps.rs page as such.

robjtede (approved):

Cargo.toml
fakeshadow: There was a UB for waker fixed in 0.3.6. So it's suggested to use at least that. It's always good to do our part on this matter.

therealprof: > Allowing versions below 0.3.7 is a security risk and shows up on our deps.rs page as such.
> 
> FYI @therealprof: https://github.com/RustSec/advisory-db/blob/master/crates/futures-util/RUSTSEC-2020-0059.md

I don't understand what you're trying to tell me, sorry. I would like to go (higher, i.e. 0.3.8), not lower.

robjtede: In terms of maintaining a library (rather than an app), allowing the most versions of crates in a dep tree makes it most compatible and gives developers the option to upgrade as they decide or, more importantly, pin a lower compatible version because of a bug/yanked crate/security issue or whatever else.

PR #454, opened 2021-01-01
remove derive-more

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Remove `derive-more` dep.
Besides the deprecated resolver actor there is only one instance of use. It's managable with manual impl of `Debug` and `Display`


src/actors/resolver.rs
```diff
@@ -121,25 +121,46 @@ impl Message for ConnectAddr {
...
+impl fmt::Debug for ResolverError {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Resolver(s) => fmt.debug_tuple("Resolver").field(s).finish(),
+            Self::InvalidInput(s) => fmt.debug_tuple("InvalidInput").field(s).finish(),
+            Self::Timeout => fmt.debug_tuple("Timeout").finish(),
+            Self::IoError(e) => fmt.debug_tuple("IoError").field(e).finish(),
+        }
+    }
+}
```
robjtede, line 147: what's up with built-in Debug derive?

src/address/mod.rs
```diff
@@ -1,7 +1,8 @@
 use std::hash::{Hash, Hasher};
-use std::{error, fmt};
-
-use derive_more::Display;
+use std::{
+    error,
+    fmt::{Debug, Display, Formatter, Result as FmtResult},
```
robjtede, line 4: nit: i prefer formatters to prefix fmt:: on usage

src/actors/resolver.rs
fakeshadow: Other error types use manual Debug so I think it could be in consistent.

robjtede (approved):
LGTM

PR #462, opened 2021-02-01
Don't require an Encoder type while constructing FramedWrite so we can be generic over the lifetime

w4:
## PR Type

Feature/Bug Fix

## PR Checklist

## Overview

**This is a breaking change as it changes the signature of `FramedWrite`**. This punts the responsibility of grabbing the type of `item` to `write()` rather than upon construction, essentially allowing us to treat the item as:

```
impl FramedWrite {
    fn write<'a>(&mut self, item: Item<'a>);
}
```

rather than the current:

```
impl<'a> FramedWrite<'a> {
    fn write(&mut self, item: Item<'a>);
}
```


robjtede:
@w4 still interested in this PR ?

PR #467, opened 2021-02-25
rework ActorFuture trait

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Rework `ActorFuture` trait. 

Move Actor type bind to trait so the type impl `ActorFuture` is not constriant to Actor type.
This enables foreign types impl `ActorFuture` without a new type. (See `actix::fut::result` and `actix::fut::ready_fut`).
And more foreign types can be introduced to further reduce new type wrappers.

### DIFF
--- a/actix/CHANGES.md
+++ b/actix/CHANGES.md
@@ -1,6 +1,11 @@
 # CHANGES
 
 ## Unreleased - 2021-xx-xx
+### Changed
+* Rework `ActorFuture` trait. [#465]
+* `actix::fut::{wrap_future, wrap_stream}` would need type annotation for Actor type. [#465]
+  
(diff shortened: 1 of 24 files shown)


robjtede:
what's the sort of user-side changes that might be needed from doing this?

fakeshadow:
Explicit annotate actor type with `actix::fut::wrap_future::<_, A>(async { })` API.
In general this can be avoid by using `async {}.into_actor(&A)`.

With a follow up change I'm planning to depercate `actix::fut::{ok, err, result, ready, Either}` and encourage users to use the std types or futures types directly.

robjtede (approved):

PR #472, opened 2021-03-01
remove ResponseChannel trait

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Remove `ResponseChannel` trait. 
It does not generic over channel type and impl only for `()` and `tokio::sync::oneshot::Sender`.
This would reduce the one generic type from `MessageResponse` trait

### DIFF
--- a/actix/src/handler.rs
+++ b/actix/src/handler.rs
@@ -1,9 +1,6 @@
-use std::fmt;
-use std::future::Future;
-use std::pin::Pin;
-use std::sync::Arc;
+use std::{fmt, future::Future, pin::Pin, sync::Arc};
 
-use tokio::sync::oneshot::Sender as SyncSender;
+pub use tokio::sync::oneshot::Sender as OneshotSender;
 
 use crate::actor::{Actor, AsyncContext};
 use crate::address::Addr;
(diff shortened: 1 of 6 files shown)


actix/src/handler.rs
```diff
@@ -1,9 +1,6 @@
-use std::fmt;
-use std::future::Future;
-use std::pin::Pin;
-use std::sync::Arc;
+use std::{fmt, future::Future, pin::Pin, sync::Arc};
 
-use tokio::sync::oneshot::Sender as SyncSender;
+pub use tokio::sync::oneshot::Sender as OneshotSender;
```
ibraheemdev, line 3: Why is this a `pub use`?

actix/src/handler.rs
```diff
@@ -602,3 +528,17 @@ SIMPLE_RESULT!(f32);
 SIMPLE_RESULT!(f64);
 SIMPLE_RESULT!(String);
 SIMPLE_RESULT!(bool);
+
+// Helper trait for send one shot message from Option<Sender> type.
+// None and error are ignored.
+trait OneshotSend<M> {
```
ibraheemdev, line 534: This trait seems a bit unnecessary.

actix/src/handler.rs
fakeshadow: actix-derive have macro for impl MessageResponse trait.

actix/src/handler.rs
fakeshadow: It's private and reduce loc. There are a lot duplicate if let in handler module

actix/src/handler.rs
robjtede: doc hidden?

fakeshadow: The derive macro would import it from actix::dev module. hide it would make people look into the expand code and wonder what is that.

fakeshadow: I think re-export the channel stuff in actix-rt could be a better way to hide all these tokio types.

robjtede (approved):

let4be:
examples are broken? https://actix.rs/book/actix/sec-2-actor.html

PR #473, opened 2021-03-02
fix earger self wakeup of actor stream

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Sytle change on handling `ActorStream` and `ActorMessageStream`.
Use relaxed yield condition for `ActorStream`.

`ActorStream`  would yield and self wake up with every ready state.
This is to not starve other actor futures from executing when actor stream yield too quickly and too much.

That said this is also an overhead on the executor as the stream keep schedule itself when the above situation happen.

This PR takes a compromise and force yield the stream only when it has 16 consecutive ready on `Stream::poll_next`. This would reduce the burden of scheduler and at the same time keep other futures waiting for a bit longer before the yield happen.

Note: 16 is a random number here with no actual reason.

### DIFF
--- a/actix/src/stream.rs
+++ b/actix/src/stream.rs
@@ -1,7 +1,7 @@
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
-use futures_core::stream::Stream;
+use futures_core::{ready, stream::Stream};
 use log::error;
 use pin_project_lite::pin_project;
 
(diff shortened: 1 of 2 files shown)


actix/src/stream.rs
```diff
@@ -129,20 +129,25 @@ where
...
+            A::handle(act, msg, ctx);
+
+            polled += 1;
+
+            if ctx.waiting() {
+                return Poll::Pending;
+            // Yield after 16 consecutive polls on this stream and self wake up.
+            // This is to prevent starvation of other actor futures when this stream yield
+            // too many item in short period of time.
+            } else if polled == 16 {
```
robjtede, line 144:
```suggestion
            } else if polled == 16 {
                // Yield after 16 consecutive polls on this stream and self wake up.
                // This is to prevent starvation of other actor futures when this stream yield
                // too many item in short period of time.
```

robjtede (approved):

PR #419, opened 2020-09-10
add ActorTryFuture{Ext} traits

robjtede:
## PR Type

Feature

## PR Checklist
- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] A changelog entry has been made for the appropriate packages.
- [x] Format code with the latest stable rustfmt

## Overview
Add `ActorTryFuture` and `ActorTryFutureExt` trait. Supports minimal combinators in the form of `map_ok`, `map_err` and `and_then`. 

Closes #314

PR #474, opened 2021-03-02
refactor actix::fut module

### LINKED ISSUE #281: ActorStream is missing some important Stream combinators
Hi,
There are some important `Stream` combinators, which are not supported by `ActorStream` right now. Especially `take_while` and `skip_while` would be nice to have access to `Actor` and `Context` instances in them. This would allow for better control over never ending streams like `Iterator` for example (e.g `take_while` would be able to access `Actor`'s field directly to check if the stream needs to be stopped).

fakeshadow:
## PR Type

Feature / Refactor

## PR Checklist

## Overview

Refactor `actix::fut` module. The goal is to make it have a similar trait sets and type names/alias to `futures` crate.

This also includes types that can be import from `futures` and impl `ActorFuture` trait directly. This would reduce dupilicate code and unify some types used between both `ActorFuture` and `Future` traits.

Add new combinators to `ActorStreamExt` trait.

This combined with https://github.com/actix/actix/pull/419 should give a rich trait sets that offers functionality close to current `futures` crate.

Closes #281


robjtede (approved):
LGTM

PR #475, opened 2021-03-03
remove IntoActorFuture trait

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Remove `IntoActorFuture` trait. All generic types bound to this trait now are bound to `ActorFuture` directly.

Rmove `WrapFuture::Output` associate type. The output type can be concrete with either input `Future::Output` type.(Or from `<WrapFuture::Future as ActorFuture<A>>::Output`).
Remove `WrapStream::Item` associate type. For similar reason as `WrapFuture`.

Refine `fut::stream` module by unify generic type alias and order. Closure types use `Fn` and `ActorFuture` types use `Fut`.

Fix error in changelog.


actix/src/fut/stream/fold.rs
```diff
@@ -5,32 +5,32 @@ use futures_core::ready;
...
 use crate::actor::Actor;
-use crate::fut::{ActorFuture, ActorStream, IntoActorFuture};
+use crate::fut::{ActorFuture, ActorStream};
 
 pin_project! {
     /// Stream for the [`fold`](super::ActorStreamExt::fold) method.
     #[derive(Debug)]
     #[must_use = "streams do nothing unless polled"]
-    pub struct Fold<S, F, Fut, T> {
+    pub struct Fold<S, Fn, Fut, T> {
```
robjtede, line 14: honestly, i prefer using F for this because Fn can confuse some parsers

fakeshadow: Ok would change it to F. My intention is to make the difference between closure and the future returned.
F would also do.

robjtede: my personal preference is F and Fut

robjtede (approved):

PR #479, opened 2021-03-07
Replace the Into implement to From<Addr> for Recipient and WeakRecipient

Folyd:
According to the [C-CONV-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#c-conv-traits) of Rust API Guidelines,
The following conversion traits should never be implemented:
- `Into`
- `TryInto`

These traits have a blanket impl based on `From` and `TryFrom`. Implement those instead.

## PR Type

Code Style

## PR Checklist


ibraheemdev (approved):
:+1: Looks good to me, is this a breaking change?

robjtede (approved):

PR #480, opened 2021-03-08
early merge on every poll of message.

### LINKED ISSUE #458: ctx.wait() does not stop processing other events sometimes
## Expected Behavior
`ctx.wait()` should always "block" actor, so it wont process any other events until waited future is resolved

## Current Behavior
`ctx.wait()` does not stop actor from processing other events sometimes, see code sample.

## Possible Solution
¯\_(ツ)_/¯

My minimal reproducible example:
```rust
use actix::prelude::*;
use actix_rt::time::sleep;
use futures::StreamExt as _;

struct MyActor {
    wait_finished: bool,
}

impl Actor for MyActor {
    type Context = Context<Self>;

        ctx.add_stream(futures::stream::iter(vec![(),(),(),()]));
    }
}

impl StreamHandler<()> for MyActor {
    fn handle(&mut self, _: (), ctx: &mut Self::Context) {
        assert!(self.wait_finished);
        self.wait_finished = false;

        ctx.wait(async {}.into_actor(self).map(|_,this,ctx| {
            let handle = ctx.run_later(std::time::Duration::from_secs(1), |_, _| {});
            ctx.cancel_future(handle);

            this.wait_finished = true;
        }));
    }
}

#[actix_rt::main]
async fn main() {
    MyActor { wait_finished: true }.start();
    sleep(std::time::Duration::from_secs(1)).await;
}
```

It is expected that this should not panic, since `handle` is supposed to run only when waited future is done.

## Your Environment

* Rust Version (I.e, output of `rustc -V`): `1.49`
* Actix Version: current master (`0.11.0-beta.1`)

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Merge context parts when there is new wait item.
This is a temporary fix that would make the related issue case work.

At the root the separation of context parts and context items is error prone and I believe it would be better to eat some cost and unify them to a single container.

This fix may also cause behavior change and potential breaking for some user code. As the merge happens early and with every poll of  concurrent message.

Closes #458

### DIFF
--- a/actix/src/contextimpl.rs
+++ b/actix/src/contextimpl.rs
@@ -4,6 +4,7 @@ use std::pin::Pin;
 use std::task::{Context, Poll};
 
 use bitflags::bitflags;
+use futures_core::ready;
 use smallvec::SmallVec;
 
 use crate::actor::{
@@ -355,12 +356,8 @@ where
             // and we always have to check most recent future
             while !this.wait.is_empty() && !this.stopping() {
                 let idx = this.wait.len() - 1;
-                if let Some(item) = this.wait.last_mut() {
-                    match Pin::new(item).poll(&mut this.act, &mut this.ctx, cx) {
-                        Poll::Ready(_) => (),
-                        Poll::Pending => return Poll::Pending,
-                    }
-                }
+                let item = this.wait.last_mut().unwrap();
+                ready!(Pin::new(item).poll(&mut this.act, &mut this.ctx, cx));
                 this.wait.remove(idx);
                 this.merge();
             }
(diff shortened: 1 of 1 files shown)


robjtede:
From a chat on discord:

> Context::spawn and Context::wait actually push a boxed future to Context parts. This is a separate container of boxed future from the real context items
Context merge is to push the futures in context parts to context item. In the linked issue case. The additional cancel trigger the cancel of run_later. But after that the logic bypass the merge and continue the next iter
Making the Context::wait future stuck in context parts until the logic decide there is no cancel to do and merge after
https://github.com/actix/actix/blob/4b6516fa2389b95e3a748bf0530c4474830cbea1/actix/src/contextimpl.rs#L292
Basically this method. In any branch we forget to merge the parts a new Context::wait could be ignored

robjtede (approved):

PR #484, opened 2021-03-22
Fix Future cancellation bug

### LINKED ISSUE #206: Future cancellation doesn't work with attached UDP stream
Using actix (d28d286) is see some behaviour that I cannot explain. My code opens a udp socket, attaches a Tokio codec and add that stream to a Actor.  Whenever a packet is received (regardless of the payload for this test) a timeout is scheduled and the corresponding `SpawnHandle` stored in a `Option`. If the `Option` is `Some` the previous timeout is canceled via `Context::cancel_future`.
I would expect that once the last packet of a sequence is received the timeout happens once. Turns out that in this scenario the timeouts aren't cancelled as expected. I couldn't reproduce the behaviour with a plain `futures::Stream::iter` - only with the UDP stuff. 

I boiled the stuff down to this test case:

```rust
extern crate actix;
extern crate futures;
extern crate tokio;

use actix::prelude::*;
use futures::prelude::*;
use std::net::SocketAddr;
use tokio::codec::BytesCodec;
use tokio::net::{UdpFramed, UdpSocket};

const ADDRESS: &str = "[::1]:42556";
const DESTINATION: &str = "[::1]:42557";

#[derive(actix::Message)]
struct Message;

#[derive(actix::Message)]
struct Timeout;

#[derive(Default)]
struct SomeActor {
    timeout: Option<SpawnHandle>,
    timeout_happened: bool,
}

impl Actor for SomeActor {
    type Context = Context<Self>;

        let addr: SocketAddr = DESTINATION.parse().unwrap();
        let sock = UdpSocket::bind(&addr).unwrap();
        let (_sink, stream) = UdpFramed::new(sock, BytesCodec::new()).split();
        ctx.add_stream(stream.map(|(_, _)| Message));

        ctx.run_later(std::time::Duration::from_secs(1), |_, _| {
            println!("Sending");
            let socket = std::net::UdpSocket::bind(ADDRESS).expect("Failed to bind address");
            for _ in 0..10 {
                socket.send_to(&[b'0'], &DESTINATION).expect("Failed to send");
            }
        });

        ctx.run_later(std::time::Duration::from_secs(3), |_, ctx| {
            ctx.stop();
        });
    }
}

impl StreamHandler<Message, std::io::Error> for SomeActor {
    fn handle(&mut self, _: Message, ctx: &mut Context<Self>) {
        if let Some(handle) = self.timeout {
            println!("Cancelling timeout");
            ctx.cancel_future(handle);
        }
        println!("Starting timeout");
        let handle = ctx.notify_later(Timeout, std::time::Duration::from_secs(1));
        self.timeout = Some(handle);
    }
}

impl Handler<Timeout> for SomeActor {
    type Result = ();

    fn handle(&mut self, _: Timeout, _: &mut Context<Self>) -> Self::Result {
        println!("Timeout! Already there? {}", self.timeout_happened);
        assert!(!self.timeout_happened);
        self.timeout_happened = true;
    }
}

#[test]
fn test_udp_stream_cancel() {
    actix::System::run(|| {
        let _ = SomeActor::default().start();
    });
}
```  

The output looks like this:

```
running 1 test
Sending
Starting timeout
Cancelling timeout
Starting timeout
Cancelling timeout
Starting timeout
Cancelling timeout
Starting timeout
Cancelling timeout
Starting timeout
Cancelling timeout
Starting timeout
Cancelling timeout
Starting timeout
Cancelling timeout
Starting timeout
Cancelling timeout
Starting timeout
Cancelling timeout
Starting timeout
Timeout! Already there? false
Timeout! Already there? true
thread 'test_udp_stream_cancel' panicked at 'assertion failed: !self.timeout_happened', tests/test_stream.rs:66:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.
Panic in Arbiter thread, shutting down system.
test test_udp_stream_cancel ... FAILED
```

What am I doing wrong? For me it looks like a bug.

evdokimovs:
## PR Type

Bug Fix

## PR Checklist

## Overview

This PR fixes #206 by removing `Item` associated with cancelled `SpawnHandle` from `ContextParts.items` if it's not found in`ContextFut.items`.

Closes #206


evdokimovs:
cc @tyranron

fakeshadow:
LGTM.

I don't see how this is an issue as by how `Context::spawn` work one should not cancel a spawn task in the same handler function. (Just don't spawn anything as what spawned would never run before the current handler yield). 

That said this is useful still for naive caller to said API and make it more dumb proof so I'm Ok with it.

fakeshadow (approved):

robjtede:
Add changelog entry please.

evdokimovs:
@robjtede ,

> Add changelog entry please.

Done.

alexlapa:
@fakeshadow ,

Here is another example that errors without this patch:

```rust
mod scheduled_task_is_cancelled_properly {
    use tokio::sync::oneshot;
    use tokio::time::timeout;

    use super::*;

    struct TestActor(Option<SpawnHandle>);

    impl Actor for TestActor {
        type Context = Context<Self>;

            // if we remove this line everything works as expected
            ctx.spawn(tokio::task::yield_now().into_actor(self));
        }
    }

    #[derive(Message)]
    #[rtype(result = "()")]
    struct Msg(oneshot::Sender<()>);

    #[derive(Message)]
    #[rtype(result = "()")]
    struct CancelTasks;

    impl Handler<Msg> for TestActor {
        type Result = ();

        fn handle(&mut self, msg: Msg, ctx: &mut Self::Context) -> Self::Result {
            let handle = ctx.run_later(Duration::from_secs(1), |_, _| {
                println!("Breakpoint B, must not print, but it prints after Breakpoint A");
                let _msg = msg;
                System::current().stop_with_code(1);
                panic!("This closure must not be executed");
            });

            self.0.replace(handle);
        }
    }

    impl Handler<CancelTasks> for TestActor {
        type Result = ();

        fn handle(&mut self, _: CancelTasks, ctx: &mut Self::Context) -> Self::Result {
            ctx.cancel_future(self.0.unwrap());
            println!("Breakpoint A, it prints before Breakpoint B");
        }
    }

    #[test]
    fn cancels() {
        let sys = System::new();
        sys.block_on(async {
            let addr = TestActor(None).start();
            let (tx, rx) = oneshot::channel();
            addr.do_send(Msg(tx));
            addr.do_send(CancelTasks);
            timeout(Duration::from_millis(2000), rx)
                .await
                .unwrap()
                .unwrap_err();
            System::current().stop();
        });
        sys.run().unwrap();
    }
}

```
Output is:
```
running 1 test
Breakpoint A, it prints before Breakpoint B
Breakpoint B, must not print, but it prints after Breakpoint A
thread 'scheduled_task_is_cancelled_properly::cancels' panicked at 'This closure must not be executed', actix/tests/test_context.rs:537:17
thread 'scheduled_task_is_cancelled_properly::cancels' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "Non-zero exit code: 1" }', actix/tests/test_context.rs:567:19
```

No canceling of spawn task in the same handler function happens here.

Notice line with `// if we remove this line everything works as expected`. I guess that this is real prerequisite for this ~~bug~~ dumb API usage.

fakeshadow:
I don't get what you mean. I already give this PR an approval.

This "patch" only fix some side effect of the `Context::spawn` API.Which is not dumb but rather how it works for now.
You would be better to avoid bad practice and work with it instead of blindly trying to treat it like `tokio::spawn`

robjtede:
released in https://github.com/actix/actix/releases/tag/v0.11.1

PR #507, opened 2021-08-01
Remove Resolver Actor

### LINKED ISSUE #451: Remove resolver actor from actix
## Expected Behavior
actix as an actor framework should not be bind to a dns dependency.
I believe majority people don't use that feature or they would be fine to move that feature to user code base.

## Current Behavior
actix was blocked from update to tokio 1.0 because the said dep is still not ready.
At this point actix is unblocked but the point still stands.

## Possible Solution
Remove the feature.

Chaostheorie:
I'm sorry for the delay. I got sidetracked and forgot to commit and push my code.

## PR Type
Refactor / Removal of deprecated feature

## PR Checklist

- [x] Documentation comments have been added / updated. (removed related docs)

## Overview
Removed actor and related docs.

Closes #451


actix/CHANGES.md
```diff
@@ -20,6 +20,9 @@
 [#499]: https://github.com/actix/actix/pull/499
 [#501]: https://github.com/actix/actix/pull/501
 
+### Removed
+- Removed `Resolver` actor [#451]
+
```
robjtede, line 25: add this to "unreleased"

robjtede (changes requested):
would also need to remove the resolver feature in cargo toml and the trust-dns deps

actix/CHANGES.md
Chaostheorie: Changes were made in new commit. Thank you for the comment

robjtede:
nice work, thanks 👍🏻

PR #524, opened 2022-02-27
implement `eq` for weak addr

### LINKED ISSUE #523: Implement PartialEq and Eq for WeakAddr
## Expected Behavior
After implementing PartialEq and Eq for `WeakAddr`  we can check if weak references point to the same actor, just as we can with their strong counterparts. Expected behavior

Let `w1`, `w2` be either a pair of `WeakAddr`s

* `w1 == w2` evaluates to `true` if and only if both point to the same actor, regardless of whether the actor reference can be upgraded
* `w1==w2` evaluates to false if and only if `w1` and `w2` reference different actors, regardless of whether the reference can be upgraded

## Current Behavior
Currently, `WeakAddr` does not allow equality comparisons.

## Possible Solution
Implement `PartialEq` and `Eq` on the type by comparing the underlying weak references to the channels.

not applicable

## Context
When in possession of only weak addresses to actors it is as useful to compare them for equality as it is for regular, strong addresses. I feel it is useful to add the equality comparisons for the sake of making the weak and strong APIs more intuitive and consistent ([principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)).

## Your Environment

* Rust Version (I.e, output of `rustc -V`): 1.56.1
* Actix Version: 0.12.0

geo-ant:
## PR Type
Feature

## PR Checklist

## Overview
We can now compare weak addresses for equality directly. In case two weak addresses `w1` and `w2` can both be upgraded, comparing `w1 == w2` will produce the same result as `w1.upgrade().unwrap() == w2.upgrade().unwrap()`. The same is true for the negation. If the weak addresses cannot be upgraded, the result still makes sense intuitively because it will check whether the weak address points to the same underlying actor, even if the actor is already destroyed.

There are no breaking changes, because `PartialEq` and `Eq` have been added to `WeakAddr` and `WeakAddressSender` and nothing else has been removed or changed.

## Issues
Closes #523

### DIFF
--- a/actix/CHANGES.md
+++ b/actix/CHANGES.md
@@ -11,7 +11,7 @@
 - Make `WeakSender` trait public rather than crate-public [#518]
 - Add `downgrade` functionality to `Recipient` to obtain a `WeakRecipient` [#518]
 - Add `From` conversion from `Recipient` to `WeakRecipient` [#518]
-
+- Add `PartialEq` to `WeakAddr` and `WeakAddressSender` [#523]
 
 ## 0.12.0 - 2021-06-06
 ### Added
--- a/actix/CHANGES.md
+++ b/actix/CHANGES.md
@@ -3,15 +3,15 @@
 ## Unreleased - 2021-xx-xx
 ### Changed
 - Updated minimum supported Rust version to 1.49.
-
-### Removed
-- Removed `Resolver` actor [#451]
 - Implement `Clone` for `WeakRecipient` [#518]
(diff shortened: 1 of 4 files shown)


robjtede (approved):
nice 👍🏻 

this will go out with 0.13 today

geo-ant:
> nice 👍🏻
> 
> this will go out with 0.13 today

thanky you kindly, this was fast!

PR #536, opened 2022-05-23
fix: ensure at least once delivery on broker IssueAsync

### LINKED ISSUE #534: Actor stops receiving messages after filling the mailbox
The following code demonstrates the bug. With actix 0.12.0 + actix-broker 1.4.1 the program prints `handle 0` to `handle 99`. With actix 0.13.0 + actix-broker 1.4.2 we only get `handle 0` to `handle 15` because after reaching mailbox capacity (16 by default) the async message subscriber gets detached from the broker.

```rust
use actix::*;
use actix_broker::*;
use std::time::Duration;

struct MyActor;

#[derive(Message, Clone)]
#[rtype("()")]
struct MyMsg(u32);

impl Actor for MyActor {
    type Context = Context<Self>;
        // ctx.set_mailbox_capacity(1_000_000_000);
        self.subscribe_async::<SystemBroker, MyMsg>(ctx);
    }
}

impl Handler<MyMsg> for MyActor {
    type Result = ();

    fn handle(&mut self, msg: MyMsg, _ctx: &mut Self::Context) -> Self::Result {
        println!("handle {}", msg.0);
    }
}

#[actix_rt::main]
async fn test() {
    let actor = MyActor;
    let _addr = actor.start();

    // give the actor time to call `started` method
    tokio::time::sleep(Duration::from_secs(1)).await;

    for i in 0..50 {
        Broker::<SystemBroker>::issue_async(MyMsg(i));
    }

    // let the actor handle all messages we sent so far
    tokio::time::sleep(Duration::from_secs(1)).await;

    for i in 50..100 {
        Broker::<SystemBroker>::issue_async(MyMsg(i));
    }
   
    // let the actor handle the second batch of messages
    tokio::time::sleep(Duration::from_secs(1)).await;
}
```

## Expected Behavior
- The actor handles all messages - 0 to 99

## Current Behavior
- Broker delivers messages until the mailbox if full
- When mailbox is full, messages are discarded
- New messages don't get delivered to the actor even when the mailbox is emptied

## Possible Solution
- You could increase size of the mailbox
- I suspect the bug was introduced here: https://github.com/actix/actix/commit/2bcb72ca267a11d9cb1e16b9f05f6bbc4174f917#diff-18ef0c609b5c921ad3b82c8a613aafce763d0458c0a945f2f0e45dee49022742R131. We could revert back to `do_send`. It inserts messages even when the mailbox is full, which could be considered a bug too, but seems to be less harmful than detaching a subscriber.

1. Create an actor with a small mailbox
2. Subscribe for some message with SystemBroker using `subscribe_async`
3. Emit at one mailbox worth of messages on the SystemBroker using `issue_async`
4. Emit some more messages after that and see if they get handled or not

## Context
I discovered the issue when a web service I have recently updated to actix 0.13.0 suddenly stopped forwarding data to other services. It did that after reaching HTTP traffic of around 100 requests per second.

## Your Environment

* Rust version (I.e, output of `rustc -V`): 1.58.1
* actix version: 0.13.0
* actix-broker version: 0.4.2

Sytten:
## PR Type

Bug fix

## PR Checklist

## Overview

While the broker API is really not great, it did offer the guarantee of at-least-once delivery previously on `issue_async`.
Since the `do_send` doesn't return a result anymore, it was changed to a `try_send` which dropped the message on mailbox full.

This PR catches this error and will enqueue the message using `do_send`. I kept the first `try_send` to detect closed mailbox and detach them from the broker (which was the previous behaviour).

Ideally the broker would follow the same API as the actor and have something like a `do_send`, `try_send` and `send` instead. But this is a larger change that is out of scope for the current fix.

Closes #534

### DIFF
--- a/actix-broker/CHANGES.md
+++ b/actix-broker/CHANGES.md
@@ -1,7 +1,7 @@
 # Changes
 
-## Unreleased - 2021-xx-xx
-
+## Unreleased - 2022-xx-xx
+- Fix at-least-once delivery guarantee for IssueAsync
 
 ## 0.4.2 - 2022-03-01
 - Add support for `actix` v0.13.
--- a/actix-broker/CHANGES.md
+++ b/actix-broker/CHANGES.md
@@ -1,7 +1,9 @@
 # Changes
 
 ## Unreleased - 2022-xx-xx
-- Fix at-least-once delivery guarantee for IssueAsync
+- Fix at-least-once delivery guarantee for IssueAsync. [#536]
+
+[#536]: https://github.com/actix/actix/pull/536
 
 ## 0.4.2 - 2022-03-01
 - Add support for `actix` v0.13.
--- a/actix-broker/src/broker.rs
+++ b/actix-broker/src/broker.rs
@@ -126,15 +126,22 @@ impl<T: 'static + Unpin, M: BrokerMsg> Handler<IssueAsync<M>> for Broker<T> {
     fn handle(&mut self, msg: IssueAsync<M>, _ctx: &mut Context<Self>) {
         trace!("Broker: Received IssueAsync");
         if let Some(mut subs) = self.take_subs::<M>() {
-            subs.drain(..)
-                .filter_map(|(id, s)| {
-                    if id == msg.1 || s.try_send(msg.0.clone()).is_ok() {
-                        Some((id, s))
-                    } else {
-                        None
+            subs.drain(..).for_each(|(id, s)| {
+                if id == msg.1 {
+                    self.add_sub::<M>(s, id);
+                } else {
+                    match s.try_send(msg.0.clone()) {
+                        Ok(_) => self.add_sub::<M>(s, id),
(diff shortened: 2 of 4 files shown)


actix-broker/CHANGES.md
```diff
@@ -1,7 +1,7 @@
 # Changes
 
-## Unreleased - 2021-xx-xx
-
+## Unreleased - 2022-xx-xx
+- Fix at-least-once delivery guarantee for IssueAsync
```
robjtede, line 4:
```suggestion
- Fix at-least-once delivery guarantee for IssueAsync. [#536]

[#536]: https://github.com/actix/actix/pull/536
```

robjtede: I would usually add this kind of thing myself but you haven't enabled "allow maintainer edits" for this PR.

robjtede (approved):

fredynaruton (approved):

PR #563, opened 2023-08-14
Add test for and fix issue with SinkWrite not sending items in buffer

otthou:
## PR Type

Bug Fix 

## PR Checklist

- [X] Tests for the changes have been added / updated.
- [X] Documentation comments have been added / updated.
- [X] A changelog entry has been made for the appropriate packages.
- [X] Format code with the latest stable rustfmt

## Overview

Currently, SinkWriteFuture only sends the first item in the buffer when polled. If the underlying Sink is Ready, and doesn't trigger the waker, and no further items are written to the SinkWrite, then the other items in the buffer are never send at all, since poll isn't called again because the task of the context isn't woken.

This can be seen in the test I added, which hangs without my fix: The Sink there doesn't block (is ready) when the messages are single bytes, so only the first byte is received on the other side of the channel. The second await hangs forever.

The fix is to simply loop and try to send all items in the buffer, until either the buffer is empty or the underlying Sink isn't ready anymore.

This fix doesn't cause any breaking changes.

### DIFF
--- a/.github/workflows/clippy-fmt.yml
+++ b/.github/workflows/clippy-fmt.yml
@@ -57,9 +57,8 @@ jobs:
       - uses: actions/checkout@v3
 
       - uses: actions-rust-lang/setup-rust-toolchain@v1.5.0
-        # temp: unpin once https://github.com/rust-lang/rust/issues/113152 is fixed
         with:
-          toolchain: nightly-2023-06-28
+          toolchain: nightly-2023-08-25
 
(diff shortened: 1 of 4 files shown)


otthou:
@robjtede Looks like the public-api-diff step is failing due to the version pinning to nightly-2023-06-28 in the workflow definition, and unrelated to my changes.

robjtede (approved):
looks good 👍🏻 thanks

robjtede:
published fix in v0.13.1

PR #604, opened 2024-12-09
fix: Forward `poll_ready()` errors to the actor and possibly finish

sdroege:
## PR Type

Bug Fix

## PR Checklist

## Overview

It's not generally supported to poll a `Sink` again after it returned an error anywhere, and e.g. `SinkMapErr` from the `futures` crate is simply panicking if this happens.

----

cc @gdesmott

### DIFF
--- a/actix/src/io.rs
+++ b/actix/src/io.rs
@@ -526,8 +526,11 @@ where
                         break;
                     }
                 }
-                Poll::Ready(Err(_err)) => {
-                    break;
+                Poll::Ready(Err(e)) => {
+                    if act.error(e, ctxt) == Running::Stop {
+                        act.finished(ctxt);
+                        return Poll::Ready(());
+                    }
                 }
                 Poll::Pending => {
                     break;


actix/src/io.rs
```diff
@@ -526,8 +526,11 @@ where
                         break;
                     }
                 }
-                Poll::Ready(Err(_err)) => {
-                    break;
+                Poll::Ready(Err(e)) => {
+                    if act.error(e, ctxt) == Running::Stop {
```
sdroege, line 530: Ideally it should not be allowed to *not* stop at this point (and below in the error cases). Polling the `Sink` again after errors is generally a bad idea.

robjtede:
Could you contrive a test that would fail on master ?

sdroege:
Yes, if this approach makes sense and you don't want to solve the underlying problem (which AFAICS requires breaking API changes)

robjtede:
To make progress, this PR needs a test and a changelog entry.

PR #624, opened 2025-05-19
Improve take_subs() function performance.

manuelgdlvh:
## PR Type
 
Refactor

## PR Checklist

### Old Behavior

Understanding the logic of Broker's i see that in the previous implementation, when broadcasting a new message, the take_subs() function removed all entries from the subscriber list, iterated over them to enqueue the message into each actor's queue, and then re-added them to the list. This approach is inefficient—not only due to the unnecessary removal and re-insertion of elements, but also because it negatively impacts data locality.

### Current Behavior

Instead of that approach, the new get_subs() function returns an iterator that casts the Any trait objects to the appropriate Recipient type for the message. This significantly improves data cache locality. Performance tests were conducted to measure the impact of this change, and the results showed notable performance gains. 

The broker performance test measures the time taken to deliver messages to varying numbers of actors—specifically 10, 25, 100, and 1000 recipients. For each configuration, the test sends 100, 1,000, and 10,000 messages, respectively, to evaluate how the system scales under different loads.

10 Actors

Screenshot from 2025-05-19 18-26-53

25 Actors

Screenshot from 2025-05-19 18-29-23

100 Actors

Screenshot from 2025-05-19 18-32-05

1000 Actors

Screenshot from 2025-05-19 18-32-36

The improvement resulted in an increase of approximately 1,500 to 2,000 messages per second.

### DIFF
--- a/actix-broker/CHANGES.md
+++ b/actix-broker/CHANGES.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Minimum supported Rust version (MSRV) is now 1.75.
+- Improve Broker message performance - take_subs() function.
 
 ## 0.4.3 - 2022-05-24
 
--- a/actix-broker/Cargo.toml
+++ b/actix-broker/Cargo.toml
@@ -21,3 +21,9 @@ log = "0.4"
 
 [dev-dependencies]
 actix-web = "4"
+criterion = { version = "0.5", features = ["html_reports"] }
+tokio = "1.44.2"
+
+[[bench]]
+name = "broker"
+harness = false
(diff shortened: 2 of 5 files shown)


robjtede:
my only concern here is that i can't run the new benchmark on macOS even when increasing the ulimit to unlimited

```
Benchmarking broker_suite_test/issue_async/Actors: 10 - Messages: 100: Collecting 10 samples in estimated 30.001 s (114k iterations)
thread 'main' panicked at /Users/rob/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/actix-rt-2.10.0/src/system.rs:42:18:
Default Actix (Tokio) runtime could not be created.: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

manuelgdlvh:
I think the problem comes from creating a new runtime in every iteration and OS not closing threads fast enough. That might be why we see too many open files error.

I'll try to reuse the same runtime each iteration instead. 

Will update you when I’ve made the change!

robjtede:
looking good, the bench runs for me now too 👍🏻

